PR #1, opened 2018-04-18
Add color primitives package

broccolini:
This pr adds primer primitives for use with Primer design system and other projects.

**To do:**
- [x] add spacing primitives
- [x] add typography primitives
- [x] setup lerna
- [x] publish color package for mobile team to test

**Some follow-up tasks:**
- [ ] use storybook or kit to output rendered version of values
- [ ] sass variables package
- [x] combined js package
- [ ] css package with custom properties ??
- [ ] Mac OS format for use in Keynote etc.
- [ ] test pulling into Figma
- [ ] test pulling into Sketch

cc @shawnbot @jonrohan


broccolini:
Added readme's and published a [primer-primitives](https://www.npmjs.com/package/primer-primitives) package that imports all the other primitives.

modules/primer-primitives/index.js
```diff
@@ -0,0 +1,10 @@
+const colors = require('primer-colors')
+const space = require('primer-spacing')
+const { fontSizes, lineHeights } = require('primer-typography')
+
+const theme = {
+  colors: colors,
+  space: space,
+  fontSizes: fontSizes,
+  lineHeights: lineHeights
+}
```
shawnbot: This could be simplified to:

```js
const theme = {colors, space, fontSizes, lineHeights}
```

Also, it looks like nothing is exported from this file. Maybe we just do this instead?

```js
module.exports = {colors, space, fontSizes, lineHeights}
```

modules/primer-typography/typography.json
```diff
@@ -8,5 +8,12 @@
     32,
     40,
     48
+  ],
+  "lineHeights": [
+    {
+      "condensedUltra": 1,
+      "condensed": 1.25,
```
shawnbot: Nit: can we sort these alphabetically?

PR #10, opened 2019-03-14
Add breakpoints.json; breakpoints and maxWidths theme keys

shawnbot:
This ports over our breakpoints from primer/components. I've made the values in the `breakpoints.json` numbers and converted them to `px` CSS strings in `breakpoints.js` because it's nice to be able to access the "raw" values without having to strip off the trailing `px`.

### TODO
- [x] Add `breakpoints.json`
- [x] Add `breakpoints.js`
- [x] Add theme `breakpoints` and `maxWidths` keys
- [ ] Change base to `master` after merging #9

### DIFF
--- a/breakpoints.js
+++ b/breakpoints.js
@@ -1,0 +1,10 @@
+const breakpoints = require('./breakpoints.json')
+const px = n => `${n}px`
+
+module.exports = {
+  breakpoints: Object.values(breakpoints).sort().map(px),
+  maxWidths: Object.entries(breakpoints).reduce((maxWidths, [key, n]) => {
+    maxWidths[key] = px(n)
+    return maxWidths
+  }, {})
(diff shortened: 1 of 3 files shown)


emplums:
@shawnbot do you want to refactor this a little bit to follow our monorepo format?

emplums:
Can you explain how the `breakpoints.json` works? We don't have anything similar for spacing so I'm curious why we need to add the `px` for our breakpoint scale and where that js file gets invoked

emplums:
Ok! It's up to you whether or not you want to push this forward or close it out. I think that we should probably spend some time thinking through what we want to do with this repo going forward since it's pretty out of date! 😬

emplums:
Same with #9! Let me know what you think!

PR #13, opened 2019-10-10
Add new spaces

### LINKED ISSUE #12: Add new spaces
Updating spaces to spaces used in the [default primer theme](https://github.com/primer/components/blob/master/src/theme.js#L98), for the theme to import the space primitive.

emersonlaurentino:
close #12 

once this is accepted import into the [default theme](https://github.com/primer/components/blob/master/src/theme.js) just like the other primitives.


simurai:
@emersonlaurentino Thanks for contributing. 🙇 

I think we currently don't use `primer/primites` in other projects. We'll include this PR in a future release once we start using it more actively.

PR #14, opened 2019-10-21
Rename package to @primer/primitives

### LINKED ISSUE #8: Rename primer-primitives to @primer/primitives
We'd like to rename this package to use the `@primer` npm scope, and move the different types of primitives into top-level imports, so:

* `primer-primitives` becomes `@primer/primitives`
* `primer-color` becomes `@primer/primitives/color`
* `primer-spacing` becomes `@primer/primitives/spacing`
* `primer-typography` becomes `@primer/primitives/typography`
* We keep (or build at publish time?) the JSON forms of each primitive type so that other languages can consume them more easily.

BinaryMuse:
This PR implements the move away from a monorepo and consolidation under `@primer/primitives` as described in #8.

- [x] Create `@primer/primitives` package at root
- [x] Remove lerna
- [x] Make old modules available at `@primer/primitives/<module>`
- [ ] Add deprecation warnings to old npm packages (do this after `@primer/primitives` is published)
- [x] Update README and other documentation as appropriate

Fixes #8


BinaryMuse:
@shawnbot I think the changes here would pretty much do it. I did have one thing I wanted to ask you about: in order to have, e.g., `@primer/primitives/typography` as a possible import, we need to either have top-level folders (`/typography`, `/spacing`, etc) at the top level of the repo, or we need to switch so that we "build" the data into a `dist/` folder that has the structure we want. I'm personally fond of not having to depend on a build step, but also not a fan of having all these loose top-level folders (and we'll need to add more for any additional categories we add). Thoughts? 💭

BinaryMuse:
I think I get where you're coming from. Although:

> 2\. The only build step I'd expect to need is one that serializes each of the endpoints as JSON for other languages to consume

Since the files are already JSON, is there anything we would need to do at all?

README.md
```diff
@@ -25,14 +16,23 @@ JSON is a highly interoperable format that can be used in many types of projects
 
 ### Example
 
-This shows an example for a React.js project. You can import Primer Primitives to provide theme values to a `ThemeProvider`. This is a great way of sharing system styles and can be achieved with popular CSS-in-JS libraries such as [styled-components](https://www.styled-components.com/) and [emotion](https://emotion.sh/).
+The Primer Primitives are exported as keys on the top-level export, or can be imported individually.
+
+```js
+import { colors, typography } from '@primer/primitives'
+// or
+import colors from '@priler/primitives/colors'
```
shawnbot: I noticed a typo here, but since this line is going away maybe it doesn't matter 😉 
```suggestion
import colors from '@primer/primitives/colors'
```

README.md
```diff
@@ -25,14 +16,23 @@ JSON is a highly interoperable format that can be used in many types of projects
...
+import typography from '@priler/primitives/typography'
```
shawnbot: Same here:
```suggestion
import typography from '@primer/primitives/typography'
```

README.md
```diff
@@ -67,14 +67,18 @@ const App = props => (
 
 ```
 
+You can also import the `colors`, `typography`, and `spacing` primitives individually if you so desire:
```
shawnbot: Sounds like this section will go away?

BinaryMuse:
Sweet — in that case, the README is updated and this should be good to go.

index.js
```diff
@@ -0,0 +1,12 @@
+const colors = require('./primitives/colors')
+const spacing = require('./primitives/spacing')
+const typography = require('./primitives/typography')
+const theme = require('./theme')
+
+module.exports = {
+  theme,
+
```
shawnbot: I don't think we need the empty line here:
```suggestion
```

emplums:
Wooooo!!! 🚢 it! 🎉

PR #21, opened 2020-10-02
Updated light.css

auareyou:
cc @simurai 

**borders**
I corrected the colors ```-secondary``` and ```-tertiary``` were pointing to. This is based on how they are used as opposed to how they map to our current css utilities. 

**states**
Replaced ```hl``` with ```state``` in the global variables section. We can keep this group for the app-level ones because some highlights aren't states but for the global ones there are only states so I figured we'd give them a more self-explanatory so they can be reused with confidence. 

**backgrounds**
I removed  ```selected: $blue-500, // `$bg-blue` in``` becuase it's already defined as ```selected-primary-bg: $blue-500,// `$bg-blue` in Primer CSS```


data/colors/light.scss
```diff
@@ -286,20 +286,20 @@ $export: (
...
     selected-primary-border: $blue-500,
   ),
 
   border: (
     primary: $gray-200,
-    primary-light: lighten($gray-200, 3%), // TODO: replace?
-    secondary: $gray-300,
-    tertiary: $gray-000,
+    primary-light: lighten($gray-200, 3%), // TODO: replace? 
+    secondary: $gray-200,
```
simurai, line 301: With this change, both `border-primary` and `border-secondary` use `gray-200`. Maybe we could do this instead:

```suggestion
    secondary: lighten($gray-200, 3%),
```

Remove `border-primary-light` since it doesn't really make sense to use the word "light" for color modes. And then use `lighten($gray-200, 3%` for `border-secondary`?

simurai (commented):
> Replaced `hl` with `state` in the global variables section.

I like this change. 😍  I think that makes it a lot easier to understand.

Added a comment about the borders.

simurai (approved):
Updated the borders e6d856bf479ff2900c7ab8b8251fb8e407da10b2 to match light.scss.

So this could be merged. 🚢 👍  Or making more changes is fine too.

PR #23, opened 2020-10-06
Update global-nav

auareyou:
Made some changes on the dark file


simurai (approved):
Maybe we should change the `ic` to `icon` to match the global variables? Otherwise 👍

PR #28, opened 2020-10-14
Update dark.scss

auareyou:
I added some annotations and change the background model slightly. I also corrected a couple of variables that were pointing to the wrong color.


simurai (approved):
Looks good. 👍 

We probably have to switch the values for `bg-secondary` <-> `bg-tertiary` in light. But can be done separately when updating Primer CSS.

PR #30, opened 2020-10-19
Update hex values in colors scales

auareyou:
Updated all of them except yellow because the yellow one isn't ready. 🥵


simurai (approved):
This should be pretty low risk since it only updates the dark colors and just changes the hex values. 🚢👍

I assume there will be a follow-up PR for the yellow scale. That maybe also involves "shifting of the scale"... like everything that used `$yellow-2` should now use `$yellow-3`. But we can look in that later.

PR #32, opened 2020-10-22
Batch of updates

auareyou:
- Corrected $white from #fff to #F0F6FC and $black to #010409
- Updated button state colors withe the final ones from @edokoa 
- Updated feedback backgrounds from the solid placeholder ones we had to the actual alphas we've created


simurai (approved):
🚢

PR #33, opened 2020-10-26
Updated diff styles

auareyou:
Add all the details for the diff component. Did it from VScode. This is better than separate PRs for each fi;e

### DIFF
--- a/data/colors/dark.scss
+++ b/data/colors/dark.scss
@@ -836,18 +836,53 @@ $export: (
   logo-subdued: $gray-6,
   discussion-border: desaturate($green-6, 40%),
 
-  diff: (
-    neutral-bg: $gray-6,
-    neutral-border: $white-fade-10,
-    deletion-bg: $red-5,
-    deletion-bg-strong: $tbd,
-    deletion-border: $red-4,
-    deletion-border-strong: $tbd,
-    addition-bg: $green-5,
-    addition-bg-strong: $tbd,
-    addition-border: $green-4,
-    addition-border-strong: $tbd,
-  ),
+diff-blob: (
+
+// empty
+
+empty-block-bg: var(--color-bg-secondary), 
+
+// deletion 
+
+    num-deletion-bg: rgba($red-5, 0.3), 
+    num-deletion-border: transparent,
+    
+    code-deletion-bg: rgba($red-5, 0.15),
+    code-deletion-border: transparent,
+    
+    marker-deletion-bg: rgba($red-5, 0.3),
+    marker-deletion-border: transparent,    
+
+// addition 
+
+    num-addition-bg: rgba($green-5, 0.3),
+    num-addition-border: transparent,
+    
+    code-addition-bg: rgba($green-5, 0.15),
+    code-addition-border: transparent,
+    
+    marker-addition-bg: rgba($green-5, 0.3),
+    marker-addition-border: transparent,
+
+// hunk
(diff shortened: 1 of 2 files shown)


auareyou:
Tried to figure out what is missing since these _are_ in both modes but still shows as failing? 

```
Variable diff-blob-code-hunk-text is missing in modes: light
Variable diff-blob-num-hunk-border is missing in modes: light
Variable diff-blob-num-hunk-icon is missing in modes: light
```

simurai:
> Tried to figure out what is missing since these _are_ in both modes but still shows as failing?

I guess using [functional CSS variables](https://github.com/primer/primitives/blob/5cad2c8fcddd2d0a46fc86f7827683e1da46575c/data/colors/light.scss#L871-L872) inside `primer/primitives` like

```scss
num-hunk-border: var(--color-border-secondary), // border-bottom
num-hunk-icon: var(--color-text-secondary),
code-hunk-text: var(--color-fade-black-70),
```

is currently not supported. What I've been doing so far is look up the value for the functional variable and use that instead:

```scss
num-hunk-border: lighten($gray-2, 3%), // border-bottom
num-hunk-icon: $gray-6,
code-hunk-text: $black-fade-70,
```

I'll change it to that now.

---

I've been wondering about this too. Also what should we do when a value of a functional variable changes. Like changing `--color-text-secondary` from `$gray-6` to `$gray-7`. I guess we would need to also go through all the component and app variables and see if those need to be updated too. That might get tricky if multiple functional variables use the same color and you can't just do a quick find+replace and have to guess which of the functional variable was meant to be used.

Here some options I can think of if we want to make it possible to use functional variables for the component and app variables:

**Option 1**: Allow `var()` as value. The output would just alias the functional variable. Not sure if that would still work for Primer (React) Components? And would need to be converted for something that doesn't use CSS, like for GitHub mobile.

```scss
num-hunk-icon: var(--color-text-secondary), // in light.scss
--color-num-hunk-icon: var(--color-text-secondary); // output
```

**Option 2**: Allow `var()` as value, but covert it to a hex value when building the theme.

```scss
num-hunk-icon: var(--color-text-secondary), // in light.scss
--color-num-hunk-icon: #586069; // output
```

**Option 3**: Similar to 2, but define the functional variables in SASS (just like the color scale) and then also include that "functional scale" it in the `$export: ()`. For component and app variables it then can be used like `num-hunk-icon: $text-secondary,`

```scss
// in light.scss
$text-secondary: $gray-6,

// add to  $export: ()
$export: (
  text-secondary: $text-secondary,

// use for component and app variables
num-hunk-icon: $text-secondary,

// output
--color-text-secondary: #586069;
--color-num-hunk-icon: #586069;
```

This would have the flexibility to make something relative to a functional variable but still slightly tweak it. E.g. `num-hunk-icon: darken($text-secondary, 3%),`. For better or worse. 😬 

**Option 4**: Keep as is. If something doesn't currently use the functional variable (`--color-text-secondary`) directly and uses its own variable (`--color-num-hunk-icon`), then it's better to not make it dependent on the functional variable even tough they currently use the same color. 

/cc @auareyou @BinaryMuse @colinkeany @broccolini for 💭 thoughts.

Personally I'm flip-flopping between 3 + 4. 🤔

simurai (approved):
🙇  Made a few tweaks, otherwise 👍 

Let's try it out. 🚢

PR #36, opened 2020-10-29
Build CSV output

colebemis:
This PR updates the build script to build CSVs. This will make it easier to import primitives into Figma.

cc @ashygee


colebemis:
No longer needed

PR #37, opened 2020-10-29
Backgrounds and borders 

auareyou:
Added ```border-overlay``` for overlay elements and changed the values in the backgrounds too. Added ```bg-backdrop``` for the fade layer for modal windows/dialogs.

### DIFF
--- a/data/colors/light.scss
+++ b/data/colors/light.scss
@@ -405,7 +405,7 @@ $export: (
     secondary: $gray-0,
     tertiary: $gray-1,
     overlay: $white,
-    backdrop: rgba($black, 0.5),
+    backdrop: $black-fade-15,
     info: $blue-0,
     info-inverse: $blue-5,
     danger: $red-0,
--- a/data/colors/light.scss
+++ b/data/colors/light.scss
@@ -388,6 +388,7 @@ $export: (
     primary: $gray-2,
     secondary: lighten($gray-2, 3%),
     tertiary: $gray-3,
+    overlay: $gray-2,
     inverse: $white,
     info: $blue-5,
     danger: $red-5,
--- a/data/colors/light.scss
+++ b/data/colors/light.scss
@@ -405,7 +405,7 @@ $export: (
     secondary: $gray-0,
     tertiary: $gray-1,
     overlay: $white,
-    backdrop: $black-fade-15,
+    backdrop: $black-fade-50,
     info: $blue-0,
     info-inverse: $blue-5,
     danger: $red-0,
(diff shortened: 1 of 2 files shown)


data/colors/light.scss
```diff
@@ -404,6 +405,7 @@ $export: (
     secondary: $gray-0,
     tertiary: $gray-1,
     overlay: $white,
+    backdrop: $black-fade-15,
```
simurai, line 408: Should this be `$black-fade-50`? Using `$black-fade-15` would be almost transparent:

`$black-fade-50` | `$black-fade-15`
--- | ---
image | Screen Shot 2020-10-30 at 10 32 43
 

```suggestion
    backdrop: $black-fade-50,
```

simurai (commented):
While looking at the [Primer CSS changes](https://github.com/primer/css/pull/1188/files) something came to mind..

Should these overlay variables be moved to the "component" section? The reason is that only Primer CSS would need these variables and they shouldn't be documented like the rest of the global variables. Hubbers shouldn't create their own "overlay components" and only use what exists (or make a feature request for a new component).

Currently Primer CSS doesn't have a component called `.Overlay`.. so this would be kinda a new category, maybe above [Components](https://github.com/primer/primitives/blob/a2fe7c0230045c83304f5a18a307e8bf4d17857d/data/colors/light.scss#L466)

```
  // Components (shared)

  overlay: (
    bg: $white,
    border: $gray-2,
    backdrop: $black-fade-15,
  ),

  // Components

  alert: (
    info: (
      text: $gray-9,
      ...
```

auareyou:
> While looking at the [Primer CSS changes](https://github.com/primer/css/pull/1188/files) something came to mind..
> 
> Should these overlay variables be moved to the "component" section? The reason is that only Primer CSS would need these variables and they shouldn't be documented like the rest of the global variables. Hubbers shouldn't create their own "overlay components" and only use what exists (or make a feature request for a new component).
> 
> Currently Primer CSS doesn't have a component called `.Overlay`.. so this would be kinda a new category, maybe above [Components](https://github.com/primer/primitives/blob/a2fe7c0230045c83304f5a18a307e8bf4d17857d/data/colors/light.scss#L466)
> 
> ```
>   // Components (shared)
> 
>   overlay: (
>     bg: $white,
>     border: $gray-2,
>     backdrop: $black-fade-15,
>   ),
> 
>   // Components
> 
>   alert: (
>     info: (
>       text: $gray-9,
>       ...
> ```

I think you're right @simurai - I tried to think of a case where it would be needed and couldn't find one. The only thing I'm wondering is if a hubber does need to build their own would we rather they at least use the right level of elevation? If they already have access to the rest of levels maybe it'd be sensitive to keep this one global too at least until we add more features to the box component or come up with a richer component for overlays. What do you think?

simurai (approved):
LGTM! 🚢👍

PR #40, opened 2020-12-15
Diff fixes dark mode

edokoa:
This is a test for the issue https://github.com/github/design-systems/issues/1168 so we can preview the changes in `lab`


edokoa:
I was testing how to do this. Going to delete the branch.

PR #45, opened 2021-03-15
Add highlight colors

simurai:
This adds the highlight colors from this [PR](https://github.com/github/github/pull/169813/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR370-R377).

Usage:

```scss
.my-class {
  color: var(--color-highlight-text);
  background: var(--color-highlight-bg);
}
```


simurai:
Merging... we can revisit again if needed.

PR #46, opened 2021-03-17
Invert tooltips in dark mode

simurai:
This changes the tooltips to be inverted in dark mode.

Before | After
--- | ---
Screen Shot 2021-03-17 at 11 08 13 | Screen Shot 2021-03-17 at 11 00 04

It solves the problem where a tooltip is used on top of another "overlay" and becomes invisible. Using a light background might be fine since tooltips are small and only visible on hover.

### DIFF
--- a/data/colors/mixins/dark_mode.scss
+++ b/data/colors/mixins/dark_mode.scss
@@ -512,8 +512,8 @@ $export: (
   ),
 
   tooltip: (
-    text: $gray-9,
-    bg: $gray-0,
+    text: $white,
+    bg: $gray-4,
   ),
 
   header-search: (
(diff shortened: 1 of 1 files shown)


josebalius (approved):
Approving from a Codespaces perspective and since I reported it 😄

auareyou:
@simurai here's the change in specs for tooltips in dark mode and dimmed mode: 
image

simurai:
> here's the change in specs for tooltips in dark mode and dimmed mode:

Ok, updated in https://github.com/primer/primitives/pull/46/commits/53fa38f6b87f2b614d69c93d35dcc6bac76786fc

PR #48, opened 2021-03-22
Update docs

colebemis:
## Problem

It's hard for consumers of the system to find and choose the correct color variables using the current docs site (https://primer.style/primitives).

## Solution

This PR makes the following changes to the docs site:

* Each column of the table can be filtered which will make it easier to search for color variables.
* Adds a column for variable type (Scale, General, Component). This gives consumers an idea of how the variable should be used.

| Before | After |
| --- | --- |
| image | image |


simurai (approved):
This is an awesome improvement! 💯  Especially that you can click on a color and it shows only the variables that use the same color. 😍

PR #49, opened 2021-03-23
Update dark toasts

simurai:
This is a follow-up to https://github.com/github/primer/issues/57#issuecomment-804044793 and updates the dark toast colors:

image


data/colors/mixins/dark_mode.scss
```diff
@@ -360,44 +360,44 @@ $export: (
...
 
     loading: (
       text: var(--color-text-primary),
-      border: var(--color-border-primary),
+      border: $gray-4,
       icon: $white,
-      icon-bg: $gray-6,
-      icon-border: var(--color-border-primary),
+      icon-bg: $gray-4,
+      icon-border: $gray-3,
```
simurai, line 400: @auareyou The `icon-border` is not defined in the spec here https://github.com/github/primer/issues/57#issuecomment-804044793, so I just used `$gray-3` (one level higher than `icon-bg`).

PR #62, opened 2021-03-29
Add `counter-secondary-bg`

simurai:
We got some feedback that the text color of `Counter--secondary` didn't have enough contrast. This PR adds `counter-secondary-bg` with opacity so it can be used for `Counter--secondary`.

Before | After
--- | ---
Screen Shot 2021-03-29 at 16 22 00 | image

The columns ☝️  above are `counter-text`, `counter-primary-text` and `counter-secondary-text`. Only the dark themes in the last column (`counter-secondary-text`) are affected.

### DIFF
--- a/.changeset/moody-lobsters-roll.md
+++ b/.changeset/moody-lobsters-roll.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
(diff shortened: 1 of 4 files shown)


auareyou (approved):
LGTM

PR #81, opened 2021-04-26
Remove blankslate-icon

simurai:
This removes `blankslate-icon`. In Primer CSS it's replaced with `icon-secondary` https://github.com/primer/css/pull/1368/commits/87a58a940790205670428ff65f307aea140935a8.

**Note**: This is a breaking change but there is no rush to merge this and we can wait for other breaking changes.


simurai:
This will be removed as part of V2.

PR #82, opened 2021-04-27
Remove border-overlay

simurai:
Closes https://github.com/github/design-systems/issues/1393.

- 🔥  Remove `border-overlay`.

⚠️ **Note**: This is a breaking change. It's not too urgent to be released. Most Primer repos already have migration PRs ready. See below.

## Migration guide

The `border-overlay` can be replaced with `border-primary`. There should be no visual difference since both variables use the same color.

- [x] Figma
- [x] https://github.com/primer/github-vscode-theme/pull/168
- [ ] https://github.com/primer/components/pull/1183
- [x] https://github.com/primer/css/pull/1370
- [x] https://github.com/primer/view_components/pull/497
- [x] https://github.com/github/github/pull/178083

### DIFF
--- a/.changeset/wild-eagles-sing.md
+++ b/.changeset/wild-eagles-sing.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": major
+---
+
+Remove border-overlay
--- a/data/colors/mixins/dark_mode.scss
+++ b/data/colors/mixins/dark_mode.scss
@@ -826,7 +826,7 @@ $export: (
 
     dropdown-text: var(--color-text-primary),
     dropdown-bg: var(--color-bg-overlay),
-    dropdown-border: var(--color-border-overlay),
+    dropdown-border: var(--color-border-primary),
     dropdown-hover-text: var(--color-state-hover-primary-text),
     dropdown-hover-bg: var(--color-state-hover-primary-bg),
     dropdown-btn-hover-text: var(--color-state-hover-primary-text),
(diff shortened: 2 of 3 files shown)


auareyou:
My suggestion was to remove `border-overlay` - It was never part of the system. Is that doable?

simurai:
> Is that doable?

Yes, doable. Just needs a few more changes in other Primer repos and on dotcom. Also, tried to figure out if we need to remove it from Figma? I found in this [Figma](https://www.figma.com/file/9vhDNyQpOono1vgi0qSViZ/Primer-Primitives%3A-Light?node-id=22%3A0) file. And it's also available from the color picker in the sidebar:

Screen Shot 2021-04-28 at 17 30 04

Not sure how/where to replace `border/overlay` with `border/primary` so that it syncs with all other Figma files that use that color. 🤔 

/cc @ashygee

auareyou:
Do you know where this one is used outside of Primer?

simurai:
> Do you know where this one is used outside of Primer?

Hard to say. We currently only keep an eye on Figma, Primer repos and dotcom. Other apps/sites will have to change it themselves when they wanna update to a newer version of Primer.

ashygee:
> Not sure how/where to replace `border/overlay` with `border/primary` so that it syncs with all other Figma files that use that color. 🤔

@simurai this would be a simple find/replace in Figma. I can show you how I do it for future updates/removals too.

simurai:
Let's close this again since in V2 `border-overlay` is mapped to `border.default`. So it will be removed anyways.

PR #85, opened 2021-05-04
Color Modes V2

simurai:
This adds a `colors_v2` folder. It currently contains only the `scale` and global v2 variables.

Below a test of the docs from https://primer.style/css/support/color-system how they might look in V2. Had to tweak it a little bit to make sure things like `--color-fg-on-emphasis` are visible:

⚠️ The screenshots below are out of date, but should give an idea the direction.

Screen Shot 2021-05-07 at 17 41 16

Screen Shot 2021-05-07 at 17 41 42


jonrohan (approved):
Looks good 👍🏻

colebemis (approved):

PR #94, opened 2021-05-14
Map v1 to v2 variables

colebemis:
Adds more mappings from v1 variables to v2 variables

Part of https://github.com/primer/primitives/issues/95


simurai:
Ok, so far by **only** using global variables we can reduce the total by half:

Screen Shot 2021-05-18 at 17 25 10

And some more once we remove `auto` colors and have replacements for transparency and fades. And maybe also when updating the color scales.

simurai:
I keep moving stuff between `light/dark_mode.scss` and `migration.scss` since I think it's not possible to have the same "key" in both files. So for now I moved most into `light/dark_mode.scss` so it's easy to compare to v1 and you can quickly uncomment something.

Issue #95, opened 2021-05-14
Map v1 variables to v2 variables


colebemis:
## Context

We will know that the [v2 color system](https://github.com/github/design-systems/issues/1388) has full coverage of the existing color system when every variable currently in Primer Primitives (v1) can be either mapped to a v2 variable or moved into github/github as an app-level variable.

## Do this

In the [`colors_v2`](https://github.com/primer/primitives/tree/main/data/colors_v2) directory of [primer/primitives](https://github.com/primer/primitives), add mappings for all v1 variables. Every v1 variable should be mapped to one of the following:
  - a v2 global variable (e.g. `var(--color-fg-default)`)
  - a v2 component variable (e.g. `var(--color-btn-primary-bg)`)
  - a scale variable (Only variables intended to be moved out of primer/primitives and into github/github can use scale variables. Indicate that a variable should be moved to github/github with a code comment)

  Use the [Migration spreadsheet](https://docs.google.com/spreadsheets/d/1vL-FPM7-NvTsHgKBRur543jnuzQJqJ69Fl6pH8EyKy4/edit#gid=752100387) as a reference.

## Example

Here's an example diff showing how to add a mapping from `text-primary` (v1 variable) to `fg-default` (v2 variable):

```diff
diff --git a/data/colors_v2/mixins/migration.scss b/data/colors_v2/mixins/migration.scss
index 9892178..a723b4a 100644
--- a/data/colors_v2/mixins/migration.scss
+++ b/data/colors_v2/mixins/migration.scss
@@ -2,3 +2,6 @@
 
 // Mapping of v1 variables to v2 variables
 
+$text: (
+  primary: var(--color-fg-default),
+);
diff --git a/data/colors_v2/mixins/dark_mode.scss b/data/colors_v2/mixins/dark_mode.scss
index f64f1b4..e55756c 100644
--- a/data/colors_v2/mixins/dark_mode.scss
+++ b/data/colors_v2/mixins/dark_mode.scss
@@ -87,4 +87,5 @@ $export: (
   ),
 
   // Mapping of v1 variables to v2 variables
+  text: $text,
 );
diff --git a/data/colors_v2/mixins/light_mode.scss b/data/colors_v2/mixins/light_mode.scss
index a5d9087..a2a411f 100644
--- a/data/colors_v2/mixins/light_mode.scss
+++ b/data/colors_v2/mixins/light_mode.scss
@@ -87,4 +87,5 @@ $export: (
   ),
   
[...]

PR #101, opened 2021-05-20
Map v1 to v2 variables (part 3)

simurai:
This continues from #94 and adds more mappings from v1 variables to v2 variables.

- [ ] ~~Figure out what to do with `fade` and `rgba()`~~ Maybe in a later PR

Part of #95


data/colors_v2/mixins/light_mode.scss
```diff
@@ -95,40 +113,39 @@ $export: (
     muted: $pink-0,
   ),
 
-  // Mapping of v1 variables to v2 variables
+  // Primer
+  // This is only meant for Primer components
 
-  text: $text,
-  icon: $icon,
+  primer: (
```
simurai, line 119: Can't really think of a good name for this group. Some alternatives:

- `misc`
- `internal`
- `components`

simurai (commented):
Ok, let's merge and test it a bit. Probably still not final.

PR #111, opened 2021-05-27
Rename calendar-day variable names

jonrohan:
This renames the calendarGraph names from `calendar-graph-day-L1-bg` to `calendar-graph-day-1-bg`. With the new output process these were outputting as `calendar-graph-day-l-1-bg`

### DIFF
--- a/.changeset/real-bottles-sniff.md
+++ b/.changeset/real-bottles-sniff.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Rename calendar-day variable names
--- a/data/colors/dark.ts
+++ b/data/colors/dark.ts
@@ -784,14 +784,14 @@ export default {
   calendarGraph: {
     dayBg: get('scale.gray.8'),
     dayBorder: 'rgba(27,31,35,0.06)',
-    dayL1Bg: '#0e4429',
-    dayL2Bg: '#006d32',
-    dayL3Bg: '#26a641',
-    dayL4Bg: '#39d353',
-    dayL4Border: 'rgba(255,255,255,0.05)',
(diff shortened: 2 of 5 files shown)


simurai (approved):
Is the `L` meant for "level"? Yeah, should be simpler without.

And we might can get away with just a `patch` since this seems VERY specific and hopefully not really used outside of dotcom.

jonrohan:
How about actually we remove them

simurai (changes requested):
The `calendarGraph` colors are [now available on dotcom](https://github.com/github/github/pull/181475/commits/77db4b41346300b28a46258a0caa38caa030fd2c). So I guess we could move them to `deprecated.ts`.

colebemis (approved):

simurai:
> If they are already on dotcom, can we just remove them from Primitives now?

Yeah, we could, but if we wanna follow semver somewhat strictly, it would need to be a major version. Is that ok? I think there are a few other variables that should be un-used.. like #82 or #81. But I closed those PRs again so that we can release them all in a bigger major version once v2 is done instead of making multiple smaller major versions that might be harder to keep track of? 🤔 

For Primer CSS @jonrohan started to test a `next_major` setup https://github.com/primer/css/pull/1402. So that it's possible to merge major changes without triggering a major release. In case we wanna consider that for Primitives too.

simurai:
> If they are already on dotcom, can we just remove them from Primitives now?

Ok, I think you're right. The "old" variable names like `calendar-graph-day-L1-bg` got already removed in one of the past few patch releases without anyone reporting it. So no need to wait for the next major release to remove the "new" variables like `calendar-graph-day-l-1-bg`. I made an alternative PR here: https://github.com/primer/primitives/pull/130

PR #112, opened 2021-05-27
Move primer variables to components

simurai:
This moves the `primer` variables from `global` to `component`. See https://github.com/github/design-systems/discussions/1429#discussioncomment-789917 for more details.

It makes them a bit more private and we don't have to document them and they also will not available as utility classes or props in PVC/PRC.


simurai:
Let's close this as stale.

PR #115, opened 2021-06-01
Auareyou changes v2

auareyou:
- Dark mode: changed opacity value in the subtle variables (except for neutral)
- Global variables in both dark and light: replaced the word 'warning' with 'attention'


colebemis (approved):
✨

PR #121, opened 2021-06-03
Deprecate auto variables

simurai:
This deprecates the `auto` variables. They will be `red`. E.g. `--color-auto-purple-6: #ff0000;`. This should prevent further uses of the auto variables.

It depends on https://github.com/github/github/pull/181958.


colebemis (approved):

colebemis:
🎉 Nice work!

PR #129, opened 2021-06-08
Deprecate fade variables

simurai:
This deprecates all the `fade` variables. They will be `red`. E.g. `--color-fade-black-15: #ff0000;`. This should prevent further use of the fade variables.

And with that, closes https://github.com/primer/primitives/issues/95

Screen Shot 2021-06-08 at 11 59 30


colebemis (approved):
🎉

PR #134, opened 2021-06-09
Remapping updates

auareyou:
- Updated alerts to use roles instead of `fg-default` - more infor [here](https://github.com/github/primer/issues/158#issuecomment-857796482)
- Rename `fg.inactive` to `fg.subtle`
- Deprecate `avatar-border`


simurai (approved):
Ok, I think this is good to go. 🚢

PR #138, opened 2021-06-11
Theme exceptions

auareyou:


data/colors_v2/themes/dark.ts
```diff
@@ -51,4 +51,15 @@ const scale = {
   pink: ['#ffdaec', '#ffbedd', '#ff9bce', '#f778ba', '#db61a2', '#bf4b8a', '#9e3670', '#7d2457', '#5e103e', '#42062a']
 }
 
+const exceptions = {
+  fg: {
+    fileIcon: get('fg.muted')
+  },
+  canvas: {
+    repoHeader: get('canvas.default')
+  },
```
colebemis, line 60: Are these mean to be global variables or product-level variables?

If they should be product-level variables, we should put these variables in `produce_light` and `product_dark`.

colebemis: Looks like @simurai already added a variable for repo headers to github/github: https://github.com/github/github/pull/182867/files

simurai (changes requested):

data/colors_v2/themes/dark.ts
auareyou: I thought that non-reusable styles that looked different across themes belong in the exceptions file. I can just close this PR then. :)

colebemis:
> I can make a similar PR for the "folder icon" since we plan to move all of the product variables to dotcom anyways.

Thank you! ❤️

data/colors_v2/themes/dark.ts
simurai: Ya, in the case of the folder icon the difference needs to be between the light theme (`scale-blue-3`) and all dark themes `fg-muted`. So it's ok to use a custom variable on dotcom.

- Needs to be different between `light` theme and `dark, dark_dimmed, dark_high_contrast` themes -> add a component, product or a custom variable on dotcom
- Needs to be different **within the same color mode**. E.g. `dark`, `dark_dimmed` or `dark_high_contrast` -> add an exception that overrides an existing variable.

PR #141, opened 2021-06-14
Border subtle → gray.5

Juliusschaeper:
Makes border.subtle same as default & muted (gray.5)


colebemis:
Should `border.subtle` be semi-transparent? I believe we use it on top of elements like buttons and avatars. cc @auareyou

auareyou:
I think it’s probably fine if it isn’t for high contrast since we don’t need it to blend. But we should keep an eye on it.

colebemis (approved):

PR #147, opened 2021-06-15
Global nav input fixes

Juliusschaeper:
Remapped inputBorder, Icon and Placeholder to gray.5 to match global border and placeholder.

### DIFF
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -125,8 +125,13 @@ const exceptions = {
       disabledBorder: alpha(get('scale.green.1'), 0.4),
       icon: get('scale.gray.9'),
       counterBg: alpha(get('scale.black'), 0.15)
+    },
+  },
+  globalNav: {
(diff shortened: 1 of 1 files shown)


data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -125,8 +125,13 @@ const exceptions = {
       disabledBorder: alpha(get('scale.green.1'), 0.4),
       icon: get('scale.gray.9'),
       counterBg: alpha(get('scale.black'), 0.15)
+    },
+  },
+  globalNav: {
+      inputBorder: get('scale.gray.5'),
```
colebemis, line 131: Should this map to `border.default` instead?

colebemis:
Where are these `global-nav` variables used? I can't find them in primer/css or github/github.

data/colors_v2/themes/dark_high_contrast.ts
Juliusschaeper: Yes, I was thinking about yesterday evening as well. I guess it’s a better way to fix this.

PR #148, opened 2021-06-15
Primary button improvements for HC

Juliusschaeper:
- Adjusted green scale to be slightly saturated
- Set border to green.2 instead of 1


data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -117,12 +117,12 @@ const exceptions = {
   btn: {
     primary: {
       text: get('scale.gray.9'),
```
colebemis, line 119: Should this be `fg.onEmphasis`?

auareyou (approved):
Makes sense to me!

PR #153, opened 2021-06-17
removed opacity from roles.muted + replaced foregrounds in diff blob

Juliusschaeper:
To increase the contrast in diffs I:
- removed opacity from roles.muted
- replaced `scale.gray.9` with `fg.onEmphasis`
- in diffBlob all light text was replaced with `fg.onEmphasis`

### DIFF
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -114,17 +114,56 @@ const exceptions = {
   neutral: {
     emphasis: get('scale.gray.6')
   },
+  accent: {
+    muted: get('scale.blue.4'),
+  },
+  success: {
+    muted: get('scale.green.4'),
+  },
+  attention: {
+    muted: get('scale.yellow.4'),
+  },
+  severe: {
+    muted: get('scale.orange.4'),
+  },
+  danger: {
+    muted: get('scale.red.4'),
+  },
+  done: {
+    muted: get('scale.purple.4'),
+  },
+  sponsors: {
+    muted: get('scale.pink.4'),
+  },
   btn: {
     primary: {
-      text: get('scale.gray.9'),
+      text: get('fg.onEmphasis'),
       border: get('scale.green.2'),
       hoverBg: get('scale.green.3'),
       hoverBorder: get('scale.green.2'),
-      disabledText: alpha(get('scale.gray.9'), 0.5),
+      disabledText: alpha(get('fg.onEmphasis'), 0.5),
       disabledBg: alpha(get('scale.green.5'), 0.6),
       disabledBorder: alpha(get('scale.green.2'), 0.4),
-      icon: get('scale.gray.9'),
+      icon: get('fg.onEmphasis'),
       counterBg: alpha(get('scale.black'), 0.15)
+    }
+  },
+  diffBlob: {
+    addition: {
+      numText: get('fg.onEmphasis')
+    },
+    deletion: {
+      numText: get('fg.onEmphasis')
+    },
+    hunk: {
+      text: get('fg.onEmphasis')
+    },
+    expander: {
+      icon: get('fg.onEmphasis'),
+      hoverIcon: get('fg.onEmphasis')
+    },
+    commentButton: {
+      icon: get('fg.onEmphasis')
     }
   }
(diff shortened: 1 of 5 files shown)


data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -114,18 +114,57 @@ const exceptions = {
...
+    hunk: {
+      text: get('fg.onEmphasis')
+    },
+    expander: {
+      icon: get('fg.onEmphasis'),
+      hoverIcon: get('fg.onEmphasis')
+    },
+    commentButton: {
+      icon: get('fg.onEmphasis')
+    }
```
simurai, line 167: This is a bit tricky. Most of these `diffBlob` variables [are deprecated](https://github.com/primer/primitives/blob/f68f10575cb3d91ecd060367670f616817996228/data/colors_v2/vars/deprecated.ts#L551-L586) and will get replaced by global variables. So I guess if we wanna have an exception only for HC, we would have to keep these variables for all the other themes?

simurai (changes requested):

data/colors_v2/themes/dark_high_contrast.ts
Juliusschaeper: Do we already have a PR where they are replaced with Global variables? If not, can we replace them and then make the exception?

simurai: > Do we already have a PR where they are replaced with Global variables?

Not yet, but as soon as we start "cleaning up", we'll remove all the deprecated variables from Primer Primitives.

> If not, can we replace them and then make the exception?

Hmm.. maaaaaybeee we could try using the "fallback mechanism" of CSS variables. Like

```scss
// only in dist/scss/colors_v2/_dark_high_contrast.scss
--color-diff-blob-expander-icon: #ffffff;

// somewhere on dotcom
color: var(--color-diff-blob-expander-icon, var(--color-fg-muted));
```

Which means that the browser will use `--color-diff-blob-expander-icon` in HC, but fall back to `--color-fg-muted` for the rest of the themes? 🤔  I would've to test if that works. It also makes things more complex and harder to guess when a variable is used where. I guess it's a difficult ⚖️  balance to keep the system as simple as possible but also be able to fine tune it for special cases.

/cc @jonrohan @colebemis for thoughts 💭 .

auareyou: All of these changes are pointing to `fg-onEmphasis`, we could create one product variable to group all of these. something like `diff.fg` or something like that? Maybe it needs a little refactoring like giving up the colored numbers (they don't pass contrast on all modes anyway) and have everything a little more simple and systematic.

colebemis: > Which means that the browser will use --color-diff-blob-expander-icon in HC since that variable is available, but fall back to --color-fg-muted for the rest of the themes?

The current Primitives setup enforces that all themes must have the same variables. Specifically, we cannot have a variable that exists in dark high contrast but not in the other themes.

We can change the Primitives setup but I actually think that rule has prevented a lot of bugs that could be hard to catch otherwise.

simurai: > We can change the Primitives setup but I actually think that rule has prevented a lot of bugs that could be hard to catch otherwise.

Yeah, I can imagine it will be a bit confusing having to track down which variables exist in which theme.

Ok, so in this case, if we want to have an exception only for HC, we would have to add again these app variables (like `--color-diff-blob-expander-icon`) to all themes. Or maybe better, try to what @auareyou suggested and reduce it to only a single app variable like `diff.fg`. I'll look into that in a separate PR.

simurai: **Update**: Using `diff.fg` for all of these variables doesn't really work since only in HC they are the same, but in other themes [they are different](https://github.com/primer/primitives/blob/f01e06f310f6fc64ca8b20d90db10bb9e30d6fed/data/colors_v2/vars/product_light.ts#L10-L19).

simurai (approved):
Did a quick test. The hover states could still be improved for the diff:

hover

But we also test this first and see how it impacts other things.

PR #154, opened 2021-06-17
Jules hc avatar and diffstat borders

Juliusschaeper:
`border.subtle` doesn’t work on Avatars and +/- diffstats since in both cases it’s too dark. Therefore I made two exceptions to keep border.subtle the same as border default and muted.

### DIFF
--- a/.changeset/little-boats-do.md
+++ b/.changeset/little-boats-do.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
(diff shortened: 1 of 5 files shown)


simurai (commented):
Same as https://github.com/primer/primitives/pull/153#discussion_r653509062. We plan to replace these variables with global variables.

simurai (approved):
Added the `avatar` and `diffstat` to the component variables so it can be overridden by HC.

PR #159, opened 2021-06-22
Updated green for primary button and success.emphasis

Juliusschaeper:
Based on early feedback and todays discussion with @broccolini we decided to 
- change success.emphasis to green.4 instead of .5
- map primary button also to the scale since colors are identical

**This change is top priority and blocks staff ship**

### DIFF
--- a/data/colors_v2/vars/component_light.ts
+++ b/data/colors_v2/vars/component_light.ts
@@ -83,7 +83,7 @@ export default {
       disabledText: alpha(get('scale.white'), 0.8),
       disabledBg: '#94d3a2',
       disabledBorder: get('border.subtle'),
-      focusBg: get('scale.green.4')',
+      focusBg: get('scale.green.4'),
       focusBorder: get('border.subtle'),
       focusShadow: (theme: any) => `0 0 0 3px ${alpha(get('btn.primary.focusBg'), 0.4)(theme)}`,
       icon: alpha(get('scale.white'), 0.8),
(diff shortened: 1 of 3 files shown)


data/colors_v2/vars/component_light.ts
```diff
@@ -83,7 +83,7 @@ export default {
       disabledText: alpha(get('scale.white'), 0.8),
       disabledBg: '#94d3a2',
       disabledBorder: get('border.subtle'),
-      focusBg: '#2ea44f',
+      focusBg: get('scale.green.4')',
```
colebemis, line 86:
```suggestion
      focusBg: get('scale.green.4'),
```

data/colors_v2/vars/component_light.ts
```diff
@@ -72,7 +72,7 @@ export default {
 
     primary: {
       text: get('scale.white'),
-      bg: '#2ea44f',
+      bg: get('scale.green.4'),
```
colebemis, line 75: Should this be `get('success.emphasis')`

colebemis (approved):

data/colors_v2/vars/component_light.ts
simurai: Hmm.. yeah, it could. I guess it depends a bit if `green.4` should be "locked in" for buttons or change along with `success.emphasis`?

Since this seems urgent, I'll merge "as is".

simurai (approved):

PR #164, opened 2021-06-23
Use default header for Box-header--blue in dark mode

simurai:
This PR un-deprecates the blue Box header variables so that it can:

- stay blue in light mode
- change to gray in dark mode

Before | After
--- | ---
Screen Shot 2021-06-23 at 16 53 32 | Screen Shot 2021-06-23 at 16 53 32
Screen Shot 2021-06-23 at 16 47 39 | Screen Shot 2021-06-23 at 16 48 41

### DIFF
--- a/.changeset/four-parrots-tap.md
+++ b/.changeset/four-parrots-tap.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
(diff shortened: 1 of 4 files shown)


simurai:
This got replaced by https://github.com/primer/primitives/pull/164.

PR #168, opened 2021-06-23
Add checks-donut

simurai:
This adds the checks-donut variables.

Screen Shot 2021-06-23 at 22 43 21

They already exist on dotcom, but we can change it there because it would change in v1. So this PR temporarily moves them here so there can be difference between v1 and v2.

## TODO

- Remove existing `checks-donut` variables on dotcom.


Juliusschaeper (approved):
Looks good to me 👍

PR #169, opened 2021-06-24
Fixed danger button hover state

Juliusschaeper:
- In HC the danger button hover text was still light instead of dark.

### DIFF
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -170,7 +170,7 @@ const exceptions = {
     danger: {
       hoverText: 'fg.onEmphasis',
       hoverIcon: 'fg.onEmphasis',
-      hoverCounterBG: 'alpha(get('scale.black'), 0.15)',
+      hoverCounterBG: alpha(get('scale.black'), 0.15),
       counterBg: alpha(get('scale.black'), 0.15)
     }
   },
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -166,6 +166,12 @@ const exceptions = {
       disabledBorder: alpha(get('scale.green.2'), 0.4),
       icon: get('fg.onEmphasis'),
       counterBg: alpha(get('scale.black'), 0.15)
+    },
+    danger: {
+      hoverText: 'fg.onEmphasis',
+      hoverIcon: 'fg.onEmphasis',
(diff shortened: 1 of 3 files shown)


data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -166,6 +166,12 @@ const exceptions = {
       disabledBorder: alpha(get('scale.green.2'), 0.4),
       icon: get('fg.onEmphasis'),
       counterBg: alpha(get('scale.black'), 0.15)
+    },
+    danger: {
+      hoverText: 'fg.onEmphasis',
+      hoverIcon: 'fg.onEmphasis',
+      hoverCounterBG: 'alpha(get('scale.black'), 0.15)',
```
simurai, line 173:
```suggestion
      hoverCounterBG: alpha(get('scale.black'), 0.15),
```

data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -166,6 +166,12 @@ const exceptions = {
       disabledBorder: alpha(get('scale.green.2'), 0.4),
       icon: get('fg.onEmphasis'),
       counterBg: alpha(get('scale.black'), 0.15)
+    },
+    danger: {
+      hoverText: 'fg.onEmphasis',
+      hoverIcon: 'fg.onEmphasis',
```
simurai, line 172: These two variables _should_ [already be dark (`scale.gray.9`)](https://github.com/primer/primitives/blob/2a77c37297673e64d0f58fefcbcfc39bb55e97dc/data/colors_v2/vars/component_dark.ts#L113). Or is this meant to avoid problems in case it changes for other dark themes?

Juliusschaeper: Hm, that’s weird, just noticed that in dark default the text is white on hover. 
Bildschirmfoto 2021-06-24 um 14 22 07

Juliusschaeper: But I think it makes sense to be consistent and add fg.onEmphasis for all hover states here.

simurai: Ohhh... that seems to be a bug. The output was:

```
--color-btn-danger-hover-text: scale.gray.9;
```

Should be fixed with https://github.com/primer/primitives/pull/169/commits/0cd799e53bc84e2b7d15a1a15396e4f8b8531fd6.

simurai (approved):

PR #179, opened 2021-06-30
Move `btn.selectedBg`

simurai:
This moves the override from `btn.primary.selectedBg` to `btn.selectedBg` and should fix https://github.com/github/github/discussions/183946#discussioncomment-916973.


Juliusschaeper (approved):
Thanks for the fix!

PR #182, opened 2021-07-20
Tooltip background change

Juliusschaeper:
- Based on internal [feedback](https://github.com/github/github/discussions/183946#discussioncomment-916598) I changed the tooltip bg from `neutral.emphasis` to `neutral.emphasisPlus`.
- For high contrast the exception to use `gray.1` was added.

**Help needed:**
_I’m not sure wether I can make an exception for a deprecated variabel 🤷‍♂️_

### DIFF
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -192,7 +192,7 @@ const exceptions = {
     },
   },
   tooltip: {
-    bg: get('scale.gray.1')
+    bg: get('scale.gray.0')
   }
 }
 
--- a/data/colors_v2/themes/dark_high_contrast.ts
+++ b/data/colors_v2/themes/dark_high_contrast.ts
@@ -125,7 +125,8 @@ const exceptions = {
     subtle: get('scale.gray.5')
   },
   neutral: {
-    emphasis: get('scale.gray.4')
+    emphasis: get('scale.gray.4'),
+    emphasisPlus: get('scale.gray.0')
   },
   accent: {
     muted: get('scale.blue.4'),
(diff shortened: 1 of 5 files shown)


data/colors_v2/themes/dark_high_contrast.ts
```diff
@@ -190,6 +190,9 @@ const exceptions = {
     expander: {
       icon: get('fg.onEmphasis'),
     },
+  },
+  tooltip: {
+    bg: get('scale.gray.1')
```
colebemis, line 195: If we need to add an exception for `tooltip.bg`, we'll have to undeprecate it by moving it back to the `vars/components_*.ts` files.

colebemis: What does `neutral.emphasisPlus` look like for tooltips in high contrast mode?

simurai: Tried to simulate it:

`scale.gray.1` | `neutral.emphasisPlus`
--- | ---
Screen Shot 2021-07-21 at 10 54 03 | Screen Shot 2021-07-21 at 10 52 57 

It's not too contrasty, but I wonder if `neutral.emphasisPlus` should be `white` or `gray-0` in high contrast? I think we don't really use `neutral.emphasisPlus` for anything yet, so might be ok to still change it?

Juliusschaeper: @simurai agree `gray-0` makes more sense here for `emphasisPlus`

simurai (commented):
Re. https://github.com/primer/primitives/pull/182/commits/20d93ef33fcb7d1f97ea12052a86a91cc2548a22

I meant we could keep `tooltip` deprecated, but override `neutral.emphasisPlus` from `scale.gray.4` to `scale.gray.0`.

simurai (approved):
Ok, I think this looks good. Let's try it out. 🚢

PR #183, opened 2021-07-21
Changed alert foregrounds to use `fg.default`

auareyou:
One of the last minute fixes we didn't get around doing before beta rollout. Reference here 👉 https://github.com/github/primer/issues/158#issuecomment-868286654


simurai (approved):
I remember someone pointing out that if the info alert has blue text it's hard to see links in it. So this should fix that. 👍

PR #184, opened 2021-07-27
Increase dark checks.inputBg contrast

simurai:
Currently the checks input is a bit hard to see in dark mode. This PR increases the contrast a bit:

Theme | Before | After
--- | ---  | ---
Dark | Screen Shot 2021-07-27 at 9 15 18 | Screen Shot 2021-07-27 at 9 15 28
Dark Dimmed | Screen Shot 2021-07-27 at 9 15 45 | Screen Shot 2021-07-27 at 9 16 03
Dark High Contrast | Screen Shot 2021-07-27 at 9 08 48 | Screen Shot 2021-07-27 at 9 09 04

An alternative might be to just use a border like the rest of the inputs, but not sure how well that would go down with @adrianmg 😅 

---

Reported in https://github.com/github/design-systems/discussions/1439#discussioncomment-879059

### DIFF
--- a/.changeset/tough-moose-unite.md
+++ b/.changeset/tough-moose-unite.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Increase dark checks.inputBg contrast
--- a/data/colors_v2/vars/product_dark.ts
+++ b/data/colors_v2/vars/product_dark.ts
@@ -91,7 +91,7 @@ export default {
     inputText: get('fg.muted'),
     inputPlaceholderText: get('fg.subtle'),
     inputFocusText: get('fg.default'),
-    inputBg: get('canvas.default'),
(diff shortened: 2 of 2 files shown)


adrianmg:
@simurai can we get a full screenshot of a real use case to judge with everything in context?

The changes suggested look good to me

simurai:
> can we get a full screenshot of a real use case to judge with everything in context?

Ahh... yes. It's this input here:

Screen Shot 2021-07-27 at 21 25 40


> The changes suggested look good to me, although `Dark Dimmed` text contrast VS the background seems worse than the previous one?

Yes, dimmed is somewhat intentionally low contrast. It still passes AA when focused:

Screen Shot 2021-07-27 at 21 33 42

jonrohan (approved):

PR #186, opened 2021-07-27
Underline nav contrast improvements

Juliusschaeper:
Based on this [feedback](https://github.com/github/github/discussions/183946#discussioncomment-1051866) I 
- Unified icon color
- Increased border hover contrast … _which was reported as well as an issue_
- ~~Mapped coral to scale~~

### DIFF
--- a/data/colors/dark.ts
+++ b/data/colors/dark.ts
@@ -601,7 +601,7 @@ export default {
   underlinenav: {
     border: alpha(get('scale.gray.6'), 0),
     borderHover: get('scale.gray.6'),
-    borderActive: get('#f78166'),
+    borderActive: '#f78166',
     text: get('scale.gray.3'),
     textHover: get('scale.gray.1'),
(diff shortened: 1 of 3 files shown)


data/colors/dark.ts
```diff
@@ -601,7 +601,7 @@ export default {
   underlinenav: {
     border: alpha(get('scale.gray.6'), 0),
     borderHover: get('scale.gray.6'),
-    borderActive: '#f78166',
+    borderActive: get('scale.coral.3'),
```
simurai, line 604: I think this needs to be changed back to the hex value since the coral scale was only added to `colors_v2` and doesn't exist in v1.

simurai (changes requested):

data/colors/dark.ts
Juliusschaeper: 👍

simurai (approved):

PR #191, opened 2021-08-02
[DO NOT MERGE] Refactored shadow styles

auareyou:
Made this change based on my conversations with @vdepizzol and this issue https://github.com/github/primer/issues/43. 
I think it's ready to be tested now. 

image

### DIFF
--- a/data/colors_v2/vars/global_dark.ts
+++ b/data/colors_v2/vars/global_dark.ts
@@ -21,7 +21,7 @@ export default {
   shadow: {
     small: '0 0 transparent',
     medium: (theme: any) => `0 3px 6px ${get('scale.black')(theme)}`,
-    large: (theme: any) => `0 8px 24px ${get('scale.black')(theme)}`,
+    large: (theme: any) => `0px 0px 0px 1px ${get('scale.gray.6')(theme)}, 0px 16px 32px ${alpha(get('scale.black'), 0.85)(theme)}, 0px 1px 3px ${alpha(get('scale.black'), 0.12)(theme)}`,
     extraLarge: (theme: any) => `0 12px 48px ${get('scale.black')(theme)}`
   },
 
--- a/.changeset/good-onions-behave.md
+++ b/.changeset/good-onions-behave.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Refactored shadow styles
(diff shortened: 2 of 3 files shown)


colebemis (approved):
💖

data/colors_v2/vars/global_dark.ts
```diff
@@ -21,7 +21,7 @@ export default {
   shadow: {
     small: '0 0 transparent',
     medium: (theme: any) => `0 3px 6px ${get('scale.black')(theme)}`,
-    large: (theme: any) => `0 8px 24px ${get('scale.black')(theme)}`,
+    large: (theme: any) => `0px 0px 0px 1px ${get('scale.gray.6')(theme)}, 0px 16px 32px ${alpha(get('scale.black'), 0.85)(theme)}, 0px 1px 3px ${alpha(get('scale.black'), 0.12)(theme)}`,
```
colebemis, line 24: This 1px gray border might conflict with modals that already have a border, right?

auareyou: I think so, this should be one of the things to test? It became a little difficult to test these in Figma. I'll let @vdepizzol chime in as well.

simurai: Tested this a bit.. the output is:

```css
--color-shadow-large: 0px 0px 0px 1px #30363d, 0px 16px 32px rgba(1,4,9,0.85), 0px 1px 3px rgba(1,4,9,0.12);
```

... and yes.. it could add a "double border" wherever a "normal" border already exists. For example the select menus already have a border and would look like:

Screen Shot 2021-08-10 at 19 53 17

I actually kinda like it.. 😆  but yeah, we would need to make sure we use the 2px border on other things too.

If we wanna keep it to 1px we would need to make sure to refactor all places that use `--color-shadow-large` and make sure it doesn't already have a border. Hmmm.. maybe out of scope for GA?

simurai: Tested this a bit.. the output is:

```css
--color-shadow-large: 0px 0px 0px 1px #30363d, 0px 16px 32px rgba(1,4,9,0.85), 0px 1px 3px rgba(1,4,9,0.12);
```

... and yes.. it could add a "double border" wherever a "normal" border already exists. For example the select menus already have a border and would look like:

Screen Shot 2021-08-10 at 19 53 17

I actually kinda like it.. 😆  but yeah, we would need to make sure we use the 2px border on other things too.

If we wanna keep it to 1px we would need to make sure to refactor all places that use `--color-shadow-large` and make sure it doesn't already have a border. Hmmm.. maybe out of scope for GA?

simurai (changes requested):

auareyou:
Going to close the PR for the second time. Seems we had one months ago but we closed it as well. Let's figure out an approach for this and then work on it.

PR #208, opened 2021-08-18
Fixed numText variables for the diff

auareyou:
I had to remove the exception for the line number in additions/deletions in the diff because it was overriding the fix we made last week.


colebemis (approved):

colebemis:
This needs a changeset, then feel free to merge @auareyou 👍

PR #224, opened 2021-08-30
Publish new functional color system

colebemis:
## Summary

For the past few months, we've been testing new color variables (stored in the `colors_v2` directory) on github.com. We're now ready to add the new color variables to the `colors` directory and deprecate old color variables. 

This PR does the following:
- Adds new functional color variables.
- Deprecates variables from the old color system
- Adds a Dark High Contrast theme

| Before | After |
| --- | --- |
| Untitled-2021-07-21-1443(3) | Untitled-2021-07-21-1443(4) |

**Note:** We should not release this change until we've updated our color system docs (https://github.com/primer/primitives/pull/225)


simurai (commented):
Should we wait with merging this until users won't be able to disable the "Dark High Contrast" feature preview?

colebemis:
@simurai Yeah, let's hold off until we remove the feature flag from github.com. In the meantime, we'll need to keep this branch up-to-date if we make bug fixes.

.github/workflows/v2-stats.yml
```diff
@@ -25,8 +25,3 @@ jobs:
         run: node .github/v2-coverage.js
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Compute variable count
-        run: npx ts-node .github/v2-var-count.ts
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
colebemis, line 32: The variables in `colors/` and `colors_v2/` are the same now so this action won't give us useful information anymore.

data/colors/dark_dimmed.ts
```diff
@@ -1,86 +0,0 @@
-import {merge} from '../../src/utils'
```
jonrohan, line 1: Does this deletion count as major? Or was the direct `data/colors/dark_dimmed` never used by external users?

data/colors/deprecations.json
```diff
@@ -1 +1,397 @@
...
+  "globalNav.icon": null,
+  "globalNav.inputBg": null,
+  "globalNav.inputBorder": null,
+  "globalNav.inputIcon": null,
+  "globalNav.inputPlaceholder": null,
+  "introShelf.gradientLeft": "accent.subtle",
+  "introShelf.gradientRight": "success.subtle",
+  "introShelf.gradientIn": "canvas.default",
+  "introShelf.gradientOut": null
+}
```
jonrohan, line 397: Just a side thought, it would be awesome to automate this script and have actions check it in on prs. So nobody will have to remember to re-run the script.

jonrohan (approved):
Overall looks pretty straight forward, had a question but not sure if I am asking something that's out of step with the plan.

data/colors/dark_dimmed.ts
colebemis: These source files are never referenced directly so it has no impact on consumers.

data/colors/deprecations.json
colebemis: Yeah, I thought about that too. But I'm curious if that's something we always want to automate.

colebemis: Let's see how the manually keep the file up-to-date goes after the v2 migration, then we can re-evaluate.

simurai (approved):

PR #233, opened 2021-09-03
Delete `data/colors_v2` directory

colebemis:
~🛑 Do not merge until dotcom no longer uses variables from the `colors_v2` directory https://github.com/github/github/pull/192017~

Now that we've officially published the [v2 functional color system](https://github.com/primer/primitives/pull/224), we can remove the experimental `colors_v2` directory that we were using to test the v2 functional color system in github.com.

| Before | After |
| --- | --- |
| Untitled-2021-07-21-1443(4) | Untitled-2021-07-21-1443(5) |


simurai (approved):
> Do not merge until dotcom no longer uses variables from the `colors_v2` directory

jonrohan (approved):
colors-v2 bundle gone from dotcom :shipit:

colebemis:
> Aren't the V2 variables still available as part of deprecated_light.ts and get output to the :root {}?

@simurai Yup. That's what allows us to remove the `colors_v2` directory without breaking anything

simurai:
> Yup. That's what allows us to remove the `colors_v2` directory without breaking anything

Hmm.. I got confused about...

> Do not merge until dotcom no longer uses variables from the `colors_v2` directory

but I think it's more about the CSS bundles importing from `colors_v2` directory and less about the v1 variables still being available. 👍

PR #238, opened 2021-09-17
Deleting deprecated dark and light colors

jonrohan:
This deletes the `deprecated_dark` and `deprecated_light` color vars from the system.

@primer/css-reviewers


simurai (approved):
LGTM! 👍

script/build.ts
```diff
@@ -83,7 +83,7 @@ async function writeModeOutput(collection: ModeCollection): Promise<void> {
 
   writeTsTypeIndex(collection)
 
-  writeDeprecations(collection)
+  // writeDeprecations(collection)
```
colebemis, line 86: We still need this function to parse the `deprecations.json` file
```suggestion
  writeDeprecations(collection)
```

data/colors/vars/deprecated_light.ts
```diff
@@ -1,583 +0,0 @@
-import {alpha, get} from '../../../src/utils'
-
-// Variables to be removed in the next major release
-
-export default {
```
colebemis, line 5: What if we had this file export an empty object instead of deleting it entirely? That way we have a place to move deprecated variables in the future.

.changeset/strong-spoons-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/primitives": major
+---
+
+Deleting deprecated dark and light colors
```
colebemis, line 5:
```suggestion
Remove deprecated color variables. See the [deprecated variables table](https://primer.style/primitives/colors#deprecated-variables) for the complete list of deprecated variables and their replacements.
```

script/build.ts
colebemis: On line 172 of this file, we assert that every deprecated variable in `deprecations.json` exists in Primer Primitives. If we want to keep deprecated variables in that file after they're removed, we'll need to remove that assertion.

jonrohan:
Closing until we can transfer memex

PR #243, opened 2021-09-22
Diff highlights + Btn primary state fixes

Juliusschaeper:
**Tldr:** Changes make diff support color blind colors and fixes issue where btn primary on hover is green instead of blue.

**Diff** - replaced hex values that point now to a step in a scale or variable for: `numBg`, `lineBg` and `wordBg`.
- `lineBg` and `wordBg` have be un-deprecated and moved to `app_light.ts` and `app_dark.ts` to match `diffBlob addition`

**Btn primary** - Replaced hex values to point to a step in a scale for: `hoveBg`, `disabledBg`

@simurai I wonder if Btn primary should be fixed in `light` instead of `light_protanopia`…?


data/colors/vars/app_dark.ts
```diff
@@ -20,6 +20,8 @@ export default {
       numText: get('fg.default'),
       fg: get('fg.default'),
       numBg: alpha(get('scale.red.4'), 0.3)
```
colebemis, line 22:
```suggestion
      numBg: alpha(get('scale.red.4'), 0.3),
```

data/colors/vars/deprecated_light.ts
```diff
@@ -542,9 +542,7 @@ export default {
       numHoverText: get('fg.default')
     },
     deletion: {
-      numHoverText: get('fg.default'),
-      lineBg: get('danger.subtle'),
-      wordBg: get('danger.muted')
```
colebemis, line 547: We should remove `deletion.lineBg` and `deletion.wordBg` from `deprecated_dark.ts` as well. We also need to make sure that they're not in the `deprecations.json`.

colebemis (dismissed):
Looks good 👍 

Just need to make sure that `diffBlog.deletion.lineBg` and `diffBlog.deletion.wordBg` are properly undeprecated.

data/colors/vars/deprecated_light.ts
simurai: Done!

simurai (commented):
> I wonder if Btn primary should be fixed in `light` instead of `light_protanopia`…?

Hmm.. I think I know what you mean:

- Use `scale.green.5` + `scale.green.2` instead of these hex values: https://github.com/primer/primitives/blob/c57b5901007313c6d451933bc0402ef280907265/data/colors/vars/component_light.ts#L90-L95
- Add those hex values as `exceptions` in https://github.com/primer/primitives/blob/c57b5901007313c6d451933bc0402ef280907265/data/colors/themes/light.ts#L65 instead.

Yeah, I think that's a good idea and should be more future proof in case we add more light themes. 👍

simurai (approved):
Ok, this looks good to me.

PR #245, opened 2021-09-23
Remove depecated color variables

colebemis:
## Changes

* All deprecated color variables have been removed. This will improve the bundle size of `@primer/primitives` and any libraries that depend on it (e.g. Primer CSS, Primer React)
*  The `dist` directory now contains a `removed` directory with data about removed variables organized by category:

  ```diff
    dist/
      js/
      ts/
      json/
      scss/
      deprecated/
  +   removed/
  +     colors.json
  ```
  Each JSON file in the `removed` directory contains a mapping of removed variables to replacement variables. Example:

  ```js
  // dist/removed/colors.json
  {
    "text.primary": "fg.default", // this means: `text.primary` is deprecated. Use `fg.default` instead
    "auto.blue.4": ["accent.fg, accent.emphasis"], // this means: `auto.blue.4` is deprecated. Use `accent.fg` or `accent.emphasis` instead
    "text.white": null // this means: `text.white` is deprecated. We don't have a replacement for it
  }
  ```

  This data will allow us to write linters to prevent usage of removed variables.

   Note: The build will fail if any variables listed in the `removed` directory still exist in the library.

* The `deprecations` directory in the `dist` directory has been renamed to `deprecated` (to be consistent with the `removed` directory):

```diff
dist/
- deprecations/
+ deprecated/
```


jonrohan (approved):
Like that you found some errors after you added the assertions. I wonder if we should eventually break that stuff off into real tests?

simurai (approved):
❤️

PR #246, opened 2021-09-24
Fix exceptions for light themes

simurai:
This is a follow-up to https://github.com/primer/primitives/pull/243.

- It outputs the exceptions added to `light_protanopia.ts`.
- Fixes the not available variables like `scale.success.0` with `scale.green.0`
- Removes the exceptions from `light.ts`, see https://github.com/primer/primitives/pull/246#discussion_r715240916.

If I got this DevTools test right, it should look like this for `light_protanopia`:

Screen Shot 2021-09-24 at 12 02 48

Although we still have to change the use of the `deletion` variables on dotcom.


data/colors/themes/light.ts
```diff
@@ -72,4 +72,4 @@ const exceptions = {
   }
 }
 
-export default merge(lightDeprecatedVars, lightAppVars, lightMarketingVars, lightComponentVars, lightGlobalVars, {scale})
```
colebemis, line 75: These light mode "exceptions" will actually propagate to all other light themes because they all extend this light theme. I think we should avoid defining `exceptions` in the base `light` and `dark` themes.

simurai: Ohh.. mean they are not really "exceptions" and more like "overrides" that will be inherited in all the other themes?

Hmm.. ok. I'll add the hex values for the primary button back to the `light` theme. Hopefully we can reduce these hex values so we don't exceptions for the rest of the themes.

/cc @Juliusschaeper

simurai: Added [a note](https://github.com/primer/primitives/pull/246/files#diff-00d50f3b99c855ce06d9098f82ed4bd2eef4a3ea42b655d99eab9ea986edcb7dR66-R69) ☝️  about not adding exceptions since this will probably come up at some point again.

colebemis (approved):
Thanks for the quick fix, @simurai!

PR #247, opened 2021-09-24
Update `fg.subtle`

simurai:
This is based on the discussion in https://github.com/github/primer/issues/292#issuecomment-925746809 and does the following:

- [x] Map the depreacated `text.disabled` to `fg.subtle` (instead of `fg.muted`).
- [x] Decrease contrast for `fg.subtle` from `gray.5` to `gray.4`.

## Ripple effect

Here the places that might need changes too:

### Line numbers

The line numbers currently use `fg-subtle`.  So if they need to pass contrast ratio, we would have to update them to `fg.muted`.

Screen Shot 2021-09-23 at 21 35 54

### DIFF
--- a/.changeset/famous-penguins-sing.md
+++ b/.changeset/famous-penguins-sing.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
+Update `fg.subtle`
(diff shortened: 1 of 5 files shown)


simurai:
Ok, never mind. According to [WCAG 2.0](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html):

> The minimum contrast success criterion (1.4.3) applies to text in the page, including **placeholder text**

Which would mean that:

- Placeholder text needs to pass contrast ratio and stay at `gray.5`
- Disabled text does not need to pass contrast ratio and can be lowered to `gray.4`

So we can't use `fg.subtle` for both.

PR #254, opened 2021-10-05
Rename color blind themes

jonrohan:
This PR renames the protanopia themes to `colorblind`


.changeset/green-apples-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/primitives": patch
```
colebemis, line 2: Isn't this a breaking change?

jonrohan:
> In Primer CSS we have that "next" thing setup so that changesets keep major changes on a separate branch.

for now I'm going to setup the next_major branch here also and we can hold off this change.

PR #256, opened 2021-10-07
Add `actionListItem` variables

auareyou:
As discussed yesterday during our pop-up session, @Juliusschaeper and I worked together to add the states in the ```actionList``` component as component-level variables in Primitives. We also added a new variable and some exceptions to make it work well in High Contrast mode. This new variable is intended to be used on the ```danger``` ```actionListItem``` for the text property. This requires a little work on the component itself @mperrotti or @pksjce, let me know if you need specs for this. 😀

cc @colebemis @tallys @vdepizzol @ashygee

### DIFF
--- a/data/colors/themes/dark_high_contrast.ts
+++ b/data/colors/themes/dark_high_contrast.ts
@@ -200,11 +200,16 @@ const exceptions = {
   },
   
   actionListItem: {
+    inlineDivider: get('border.default'),
+    
     default: {
-      hoverBg: 'rgba(201, 206, 212, 0.12)'
+      hoverBg: alpha(get('scale.gray.2'), 0.12),
+      activeBg: alpha(get('scale.gray.2'), 0.24),
+      selectedBg: alpha(get('scale.gray.2'), 0.08),
     },
     danger: {
-      hoverBg: get('danger.muted'),
+      hoverBg: get('danger.emphasis'),
+      activeBg: get('scale.red.6'),
       hoverText: get('fg.onEmphasis')
     }
   }


data/colors/themes/dark_high_contrast.ts
```diff
@@ -198,6 +198,18 @@ const exceptions = {
...
     },
   },
+  
+  actionListState: {
+    default: {
+      hover: get('rgba(201, 206, 212, 0.12)'),
+      focus: get('rgba(201, 206, 212, 0.24)')
+    },
+    danger: {
+      hover: get('danger.muted'),
```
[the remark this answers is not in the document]
auareyou: Yes, makes total sense!

data/colors/vars/component_dark.ts
```diff
@@ -146,5 +146,17 @@ export default {
   underlinenav: {
     icon: get('fg.subtle'),
     borderHover: get('neutral.muted')
+  },
+
+  actionListState: {
```
colebemis, line 151: IMO `actionListItem` is a little more clear than `actionListState`:

```suggestion
  actionListItem: {
```

data/colors/themes/dark_high_contrast.ts
```diff
@@ -198,6 +198,18 @@ const exceptions = {
       numBg: alpha(get('accent.muted'), 0.4),
     },
   },
+  
+  actionListState: {
```
colebemis, line 202:
```suggestion
  actionListItem: {
```

.changeset/strange-rats-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
+Add `actionListState` variables
```
colebemis, line 5:
```suggestion
Add `actionListItem` variables
```

data/colors/themes/dark_high_contrast.ts
```diff
@@ -198,6 +198,18 @@ const exceptions = {
...
+    default: {
+      hoverBg: 'rgba(201, 206, 212, 0.12)',
+      focusBg: 'rgba(201, 206, 212, 0.24)'
+    },
+    danger: {
+      hoverBg: get('danger.muted'),
+      hoverText: get('fg.onEmphasis'),
```
vdepizzol, line 209: @auareyou @Juliusschaeper do we have ways to validate how this looks in high contrast? Here's what I'm seeing this when testing these values: ☹️

Screen Shot 2021-10-08 at 16 20 21

Also, what's your recommended approach for interactive states in high contrast? Do the `:hover`, `:focus` styles become solid? As opposed to some of our current components (that show a blue background with white text), `actionList` doesn't invert the text color since it can support a lot of rich content inside. But I'm not sure if that's something we should do in high contrast.

data/colors/vars/component_dark.ts
```diff
@@ -146,5 +146,17 @@ export default {
...
+  actionListItem: {
+    default: {
+      hoverBg: 'rgba(201, 206, 212, 0.12)',
+      focusBg: 'rgba(201, 206, 212, 0.24)'
+    },
+    danger: {
+      hoverBg: 'rgba(248, 81, 73, 0.16)',
+      hoverText: get('danger.fg'),
+      focusBg: 'rgba(248, 81, 73, 0.16)'
+    }
```
vdepizzol, line 160:
```suggestion
  actionListItem: {
    inlineDivider: alpha(get('border.default'), 0.48),
  
    default: {
      hoverBg: alpha(get('scale.gray.2'), 0.12),
      activeBg: alpha(get('scale.gray.2'), 0.2),
      selectedBg: alpha(get('scale.gray.2'), 0.08),
    },
    danger: {
      hoverBg: alpha(get('scale.red.4'), 0.16),
      activeBg: alpha(get('scale.red.4'), 0.24),
      hoverText: get('scale.red.3')
    }
```

data/colors/vars/component_light.ts
```diff
@@ -147,5 +147,20 @@ export default {
...
+  actionListItem: {
+    inlineDivider: alpha(get('border.default'), 0.48),
+    
+    default: {
+      hoverBg: alpha(get('scale.gray.2'), 0.32),
+      activeBg: alpha(get('scale.gray.2'), 0.48),
+      selectedBg: alpha(get('scale.gray.2'), 0.24)
+    },
+    danger: {
+      hoverBg: alpha(get('danger.subtle'), 0.64),
```
colebemis, line 161: Is there a reason the `light` theme uses `danger.*` variables and the `dark` theme uses `scale.red.*` variables?

data/colors/vars/component_dark.ts
```diff
@@ -146,5 +146,20 @@ export default {
...
     borderHover: get('neutral.muted')
+  },
+
+  actionListItem: {
+    inlineDivider: alpha(get('border.default'), 0.48),
+  
+    default: {
+      hoverBg: alpha(get('scale.gray.2'), 0.12),
+      activeBg: alpha(get('scale.gray.2'), 0.2),
+      selectedBg: alpha(get('scale.gray.2'), 0.08),
```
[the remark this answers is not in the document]
vdepizzol: nope :))

data/colors/vars/component_light.ts
vdepizzol: @colebemis it looks like `danger.subtle` on `light` theme uses a fully opaque light red, while `danger.subtle` on dark mode uses a 15% opacity red. Would `alpha()` on a color with transparency like that  change its current opacity to a different value?

data/colors/themes/dark_high_contrast.ts
Juliusschaeper: @vdepizzol and me met to discuss HC states for the action list:

- For the default item we focused on having a clearly differentiable outline while keeping the fill as it is to allow for rich content. 
- The danger item follows the general HC approach of using emphasis + fg.onEmphasis.
- _Now looking at it, I wonder if the .selected state should also use emphasis + fg.onEmphasis 🤔_

Dark HC Neutral
Dark HC Danger

PR #259, opened 2021-10-21
[experiment][do not merge] shading a canvas or UI plane

mperrotti:
These changes add partially transparent colors that will have some contrast with any color they are placed on top of.

For example:
The ActionList component does not have a background color applied to it, so its background color depends on where it is placed. The items in an ActionList can use these colors as background colors for `:hover`, `:focus`, `:active`, and `selected` styles to ensure the state will be visible no matter what background color the component appears on.

Draft PR using these colors in react/primer: https://github.com/primer/react/pull/1527

### DIFF
--- a/data/colors/vars/global_dark.ts
+++ b/data/colors/vars/global_dark.ts
@@ -18,6 +18,11 @@ export default {
     muted: get('scale.gray.7'),
     subtle: alpha(get('scale.gray.0'), 0.1)
   },
+  shade: {
+    default: alpha(get('scale.gray.2'), 0.2),
+    muted: alpha(get('scale.gray.2'), 0.12),
+    subtle: alpha(get('scale.gray.2'), 0.08),
+  },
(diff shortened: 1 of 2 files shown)


mperrotti:
Thanks for the great feedback @simurai !

I'm leaning towards `primer.state.{state}.bg` because `primer.canvas.{state}` makes it seem like it's closely related to the `canvas` colors.

@vdepizzol and I are going to talk about this during our weekly design pattern working session.

simurai:
> because `primer.canvas.{state}` makes it seem like it's closely related to the `canvas` colors.

Would these new background colors also work for let's say on top of `accent.subtle`? If so, then yeah.. calling it `primer.canvas.{state}` would make that confusing.

> I'm leaning towards `primer.state.{state}.bg`

Yeah, I think it would be nice that all "state" colors are grouped together and not spread apart by "property". 👍

PR #265, opened 2021-11-01
Prepare for new gloss buttons with color mode support

tobiasahlin:
This PR updates color variables to replace the temporary overrides in https://github.com/primer/css/pull/1716 👀

### DIFF
--- a/.changeset/swift-berries-learn.md
+++ b/.changeset/swift-berries-learn.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": major
+---
+
+Prepare for new gloss buttons with color mode support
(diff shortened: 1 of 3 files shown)


simurai (commented):
A question: Are these variables meant for all themes or only the `light` and `dark`, but not `dark_dimmed`, `dark_high_contrast` etc.

The reason for asking is that using hex values like '#f6f8fa' will be the same in all dark themes. Is that ok?

tobiasahlin:
@simurai Yesss, right now that's the plan 🙏 we don't have any other variants defined or in the roadmap, and these variants should work well in both `dark` and `dark_dimmed`, etc (contrast wise)

simurai (approved):
> Yesss, right now that's the plan 🙏 we don't have any other variants defined or in the roadmap, and these variants should work well in both `dark` and `dark_dimmed`, etc (contrast wise)

Ok, great. Diff looks a lot simpler now. ✌️

PR #274, opened 2021-11-18
Improved contrast for change highlights in diff

Juliusschaeper:
wordBg → emphasis
word foreground → onEmphasis

This increases contrast for change highlights to background and foreground.

### DIFF
--- a/.changeset/wet-geese-obey.md
+++ b/.changeset/wet-geese-obey.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Improved contrast for change highlights in diff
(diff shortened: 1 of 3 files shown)


simurai (commented):
@Juliusschaeper Is this PR meant to change also the diff for the `Light default` theme?

If it's meant only for `Light high contrast`, then I guess it would need to be an [exception](https://github.com/primer/primitives/blob/38a8e872896a952914e0495cc6f2252eaa7975db/data/colors/themes/light_high_contrast.ts#L117).

Juliusschaeper:
@simurai Thanks for noticing this issue 🙈 I reverted my changes and moved the adjustments over to light high contrast exceptions.

simurai (approved):

PR #281, opened 2021-12-01
Improve diff-blob-expander icon (follow-up)

simurai:
Follow-up to https://github.com/primer/primitives/pull/280.

For LHC, we actually want to change the background and not the icon color, see https://github.com/github/design-infrastructure/discussions/1439#discussioncomment-1721488. So this PR:

- [x] Reverts changing the icon color https://github.com/primer/primitives/commit/a65bb0f3b2e09cba38ad5dece108a20be3dc0d87
- [x] But use `fg.default` for the icon in DHC https://github.com/primer/primitives/pull/281/commits/7512227e9ef9db7c126866d672a4fa41f9188063 because we will be using `--color-diff-blob-hunk-num-bg` on [dotcom](https://github.com/github/github/pull/202115) which is quite dark.

### DIFF
--- a/.changeset/metal-kids-smile.md
+++ b/.changeset/metal-kids-smile.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Improve diff-blob-expander-icon (follow-up)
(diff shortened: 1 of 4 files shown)


siddharthkp (approved):
Makes a lot of sense!

Juliusschaeper (approved):

PR #283, opened 2021-12-14
Background adjustments

Juliusschaeper:
Adjusted backgrounds for consistent states in Dark and Light high contrast. 
#### Next step: 
- Add borders to the component so we can enable them in high contrast. Here are the specs [Figam](https://www.figma.com/file/ekYFyHK592wxLw1pxhDGEf/ActionList-colors?node-id=62%3A1179)

| Default | High contrast |
|--------|--------|
| Default | High contrast |

### DIFF
--- a/.changeset/popular-cameras-drive.md
+++ b/.changeset/popular-cameras-drive.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
+Background adjustments for High Contrast theme.
--- a/data/colors/themes/dark_high_contrast.ts
+++ b/data/colors/themes/dark_high_contrast.ts
@@ -204,8 +204,10 @@ const exceptions = {
     
     default: {
       hoverBg: get('scale.gray.8'),
+      hoverBorder: get('border.default'),
       activeBg: get('scale.gray.6'),
-      selectedBg: get('scale.gray.6'),
+      selectedBg: get('scale.gray.6'), 
+      selectedBorder: get('scale.gray.4')
     },
     danger: {
       hoverBg: get('danger.emphasis'),
--- a/data/colors/themes/dark_high_contrast.ts
+++ b/data/colors/themes/dark_high_contrast.ts
@@ -198,16 +198,16 @@ const exceptions = {
       numBg: alpha(get('accent.muted'), 0.4),
     },
   },
-  
+
   actionListItem: {
     inlineDivider: get('border.muted'),
-    
+
     default: {
       hoverBg: get('scale.gray.8'),
       hoverBorder: get('border.default'),
       activeBg: get('scale.gray.6'),
-      selectedBg: get('scale.gray.6'), 
-      selectedBorder: get('scale.gray.4')
+      activeBorder: get('scale.gray.4'),
+      selectedBg: get('scale.gray.6')
     },
     danger: {
       hoverBg: get('danger.emphasis'),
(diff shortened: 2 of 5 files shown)


langermank:
@Juliusschaeper I deployed this to a draft PR so we can see how it looks in Storybook: https://primer-css-bs455yyc6-primer.vercel.app/css/storybook?path=/story/components-actionlist-patterns--action-list-full-bleed&globals=theme:all

PR here if you're curious: https://github.com/primer/css/pull/1830/files

For the borders, we can also test that in the draft PR. What tokens would we use for that? From what I can tell you haven't added that here yet but I may have missed it. We would need the value to be defined for all themes, even if its not "visible"- so for the light theme, the value would just match that of the item background color.

simurai:
> For the borders, we can also test that in the draft PR. What tokens would we use for that?

We could do the same as for the `Counter`. Add a `actionListItem.default.hoverBorder` that is [`transparent`](https://github.com/primer/primitives/blob/bec27f1c0384f6193f628e5766e35ab220ec1c76/data/colors/vars/component_light.ts#L11-L13) for light and dark mode and then add [an exception](https://github.com/primer/primitives/blob/bec27f1c0384f6193f628e5766e35ab220ec1c76/data/colors/themes/light_high_contrast.ts#L159-L161) for the "dark/light high contrast" themes.

Side note: In https://github.com/primer/primitives/pull/259 we're still thinking of having `primer.state` variables, so that we might can share the same state variables across multiple components. But for this PR adding the borders to [`actionListItem`](https://github.com/primer/primitives/blob/bec27f1c0384f6193f628e5766e35ab220ec1c76/data/colors/vars/component_light.ts#L156) should be fine.

simurai (approved):
Ok, with https://github.com/primer/primitives/pull/283/commits/b326cc0c0c14badc4f8a38a826aa050c4de581f0 added,  I think this PR is ready. 🚢 

We still have to use the new variables in PRC and PCSS.

langermank:
@simurai @Juliusschaeper pushed another test to Primer CSS, here's how the borders look (sorry I kind of broke the theme viewer recently 🙄) https://primer-css-rnc8wylnl-primer.vercel.app/css/storybook?path=/story/components-actionlist-actionlistitem--playground&globals=theme:all

simurai (commented):
@Juliusschaeper correct me if I'm wrong, but I think the border/outline should only appear on `hover` and `active` state? Like this:

Screen Shot 2021-12-21 at 17 21 27

So I assume:

- Add `--color-action-list-item-default-hover-border` to [:hover](https://github.com/primer/css/blob/47a57b36f0cc08ea9863911a3603e22f7f0b5780/src/actionlist/action-list-item.scss#L39)
- Add `--color-action-list-item-default-selected-border` to [:active](https://github.com/primer/css/blob/47a57b36f0cc08ea9863911a3603e22f7f0b5780/src/actionlist/action-list-item.scss#L44)

But if so, I guess we should rename `selectedBorder` -> `activeBorder`. Since the selected state has no border. Push that change directly to this PR.

Juliusschaeper:
@simurai yes that's correct. The selected states already has a strong signifier.

simurai:
This change is currently trickling down to PCSS in https://github.com/primer/css/pull/1856.

/cc @primer/react-reviewers for awareness to make the same change to PRC as well.

mperrotti:
Thanks for tagging us @simurai 
Issue created: https://github.com/github/design-infrastructure/issues/2244

PR #288, opened 2022-01-26
Add utility mapping

simurai:
This adds the [Primer CSS utilities](https://github.com/primer/css/blob/fdacefdbf5d8d7028b7147ce666c8f568dc5d5ea/src/utilities/colors.scss) mapped to Primitives. As of now, this will only be used by the Figma plugin to import the colors since Figma also uses the "utility" names.

### Open questions

- Do we even want this "utility" mapping in Primitives? It's hard to explain why they are needed. A longer term solution might be to add the utilities as [global variables](https://github.com/primer/primitives/blob/main/data/colors/vars/global_light.ts). It's a bigger effort since we probably also want to remove some of the existing variables so they are not unnecessarily duplicated. Like we probably don't wanna keep both `bg.default` AND `canvas.default`.


colebemis (approved):
I agree that these utilities are hard to explain and I'd like to avoid people using these directly outside of Figma. This implementation seems like a good way to support Figma while keeping these utilities "private."

PR #301, opened 2022-03-01
[Actions] Update light theme color for gateWaitingText

konradpabjan:
Closes: https://github.com/github/c2c-actions-checks/issues/242

Currently the warning text color is incorrect but only in the light theme:

This will update the warning text to match warnings in logs:

## Current incorrect color

image

## With updated color

image

In our codebase we have `color: var(--color-checks-gate-waiting-text);` 

image

This matches the same color we have for warnings in our logs for the light theme

This is exactly what `stepWarningText: get('scale.yellow.3'),` is

image

### DIFF
--- a/.changeset/eleven-badgers-divide.md
+++ b/.changeset/eleven-badgers-divide.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+[Actions] Update light theme color for gateWaitingText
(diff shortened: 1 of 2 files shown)


simurai (approved):
This makes sense to me. 👍 

@konradpabjan thanks for the fix. 🙇

PR #304, opened 2022-03-04
Updates white in dark.ts to the actual white color

edokoa:
This is required to make `fg-on-emphasis` work on `emphasis-plus` [Tooltips need this to pass AA contrast]

### DIFF
--- a/.changeset/spotty-seahorses-pay.md
+++ b/.changeset/spotty-seahorses-pay.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Change `scale.white` in dark mode to `#ffffff`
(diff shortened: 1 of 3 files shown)


edokoa:
This comes from this discussion in Special Projects about tooltips and accessibility.

* https://github.com/github/special-projects/issues/574

After talking with @Juliusschaeper about the color scales, we decided to correct `white` in normal contrast dark themes to `#ffffff` so the color is displayed correctly in Tooltips and passes AA contrast requirements.

simurai (approved):

simurai:
That should also improve `neutral-emphasis` since it's the same as `neutral-emphasis-plus`:

Screen Shot 2022-03-04 at 20 14 53

PR #309, opened 2022-03-21
Add component primitives for the toggle switch component

mperrotti:
Will be used for [this primer/react PR](https://github.com/primer/react/pull/1933)

### DIFF
--- a/data/colors/vars/component_dark.ts
+++ b/data/colors/vars/component_dark.ts
@@ -167,5 +167,15 @@ export default {
       activeBg: alpha(get('scale.red.4'), 0.24),
       hoverText: get('scale.red.3')
     }
-  }
+  },
+
+  switchTrack: {
+    bg: get('scale.black'),
+
+    checked: {
+      bg: alpha(get('scale.blue.5'), 0.35),
+      hoverBg: alpha(get('scale.blue.5'), 0.5),
+      activeBg: alpha(get('scale.blue.5'), 0.65),
+    }
+  },
 }
--- a/data/colors/vars/component_light.ts
+++ b/data/colors/vars/component_light.ts
@@ -168,5 +168,24 @@ export default {
       activeBg: get('danger.subtle'),
       hoverText: get('danger.fg')
     }
+  },
+
+  switchTrack: {
+    bg: get('scale.gray.1'),
+    border: get('scale.gray.3'),
+
+    checked: {
+      bg: get('scale.blue.0'),
+      hoverBg: get('scale.blue.1'),
+      activeBg: get('scale.blue.2'),
+      border: get('scale.blue.3'),
+    }
+  },
+
+  switchKnob: {
+    checked: {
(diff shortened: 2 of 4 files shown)


data/colors/vars/component_light.ts
```diff
@@ -168,5 +168,24 @@ export default {
...
+
+    checked: {
+      bg: get('scale.blue.0'),
+      hoverBg: get('scale.blue.1'),
+      activeBg: get('scale.blue.2'),
+      border: get('scale.blue.3'),
+    }
+  },
+
+  switchKnob: {
```
siddharthkp, line 185: Curious: Noticed that `switchKnob` is component_light and not in the others. How do we structure these?

data/colors/vars/component_dark.ts
```diff
@@ -167,5 +167,24 @@ export default {
...
+
+    checked: {
+      bg: alpha(get('scale.blue.5'), 0.35),
+      hoverBg: alpha(get('scale.blue.5'), 0.5),
+      activeBg: alpha(get('scale.blue.5'), 0.65),
+      border: get('scale.blue.3'),
+    }
+  },
+
+  switchKnob: {
```
siddharthkp, line 184: Curious: How do we decide the nesting structure in primitives, are there any rules/ADRs around it?

switch: {
  track: {},
  knob: {}
}

vs 

switchTrack: {},
switchKnob: {}

mperrotti: I'm not aware of any rules around this. I considered nesting `switch.track` and `switch.knob` too, but decided to keep it flatter.

I don't feel strongly either way.

data/colors/vars/component_dark.ts
```diff
@@ -167,5 +167,24 @@ export default {
       activeBg: alpha(get('scale.red.4'), 0.24),
       hoverText: get('scale.red.3')
     }
+  },
+
+  switchTrack: {
+    bg: get('scale.black'),
+    border: get('scale.gray.4'),
```
langermank, line 174: Is this new border color one that we want to share with _all_ form controls? Like `input` and `select`? Thinking about the contrast ratio for our standard `input` border..

mperrotti: I was worried about making that decision in the scope of this PR. These changes just support the ToggleSwitch component, and I think we'll want to zoom out and look at all form inputs.

Here's what I proposed to Allie and Chelsea:
Screen Shot 2022-03-21 at 4 17 53 PM

langermank: Thanks for that context! That makes sense to me as well. If you add a more generic token _now_ without actually rolling it out to more form components or documenting it, I could see that being confusing for folks utilizing tokens in other apps. Renaming and re-mapping tokens later isn't a huge lift 😄

siddharthkp (approved):

PR #310, opened 2022-03-25
Add Tritanopia theme

simurai:
This is in support of https://github.com/github/primer/issues/844 and adds 2 new `light_tritanopia` and `dark_tritanopia` themes.

They are identical to the `colorblind` themes except for:

- `orange` scale use the values from the default `red` scale
- `red` scale use the values from the default `red` scale

Light colorblind | Light Tritanopia ✨  new
--- | ---
Screen Shot 2022-03-25 at 15 14 02 | Screen Shot 2022-03-25 at 15 13 31

Here the effect on the open `State` component:

Light colorblind | Light Tritanopia ✨  new
--- | ---
Screen Shot 2022-03-24 at 18 23 40 | Screen Shot 2022-03-24 at 18 23 55

And here a [simulation](https://www.color-blindness.com/coblis-color-blindness-simulator/) how a user with Tritanopia might see the open `State`. It looks "red" in both themes but a bit more vibrant in the Tritanopia theme:

Light colorblind | Light Tritanopia ✨  new
--- | ---
dp-theme | t-theme
Open = red | Open = red (more vibrant)


data/colors/themes/light_tritanopia.ts
```diff
@@ -0,0 +1,62 @@
+import {merge} from '../../../src/utils'
+import light_colorblind from './light_colorblind'
```
simurai, line 2: Is it ok to import `light_colorblind` directly (instead of `light`)? It would avoid having to keep the [colorblind exceptions](https://github.com/primer/primitives/blob/251de59997a90bb9500a87152fcbee99798ae3e8/data/colors/themes/light_colorblind.ts#L62-L93) in sync.

lesliecdubs:
@jonrohan do you have capacity to hop in on this review this week to help get color modes to beta?

PR #313, opened 2022-04-06
add custom formats for commonjs modules and typescript

rezrah:
### Summary

Adds two custom `StyleDictionary` formats for `commonjs` and `typescript`, which enable parity with existing Primitive object notation conventions.

### Why make this change?

1. Default `javascript/module` format in `StyleDictionary` outputs nested JSON, which includes a value key. This would - if left unchanged - break existing conventions and isn't ideal for Primitive consumers like PRC that require the object-notation access to be sans-value.

    E.g. 
    
    **Default**: `gh.text.title.size.large.value`
    **What it should be**: `gh.text.title.size.large`

2. The new `shorthand` value used in `typography.json` departs from StyleDictionary syntax conventions by not providing an object with a `value` key. 

    E.g.
    Screenshot 2022-04-06 at 10 31 00
    
    Default StyleDictionary formats require the `value` key to transform correctly. As it was missing, these were being omitted from the output.

3. Default commonjs formats include the `DesignToken` interface as the value, which leads to a unnecessarily bloated output and heavier payload for the client.
    Screenshot 2022-04-06 at 10 38 24

4. The default type definition for token values is `any` (see above). This should be more specific.

#### CommonJS format
<table>
<tr>
<th> Before </th> <th> After </th>
</tr>
<tr>
<td valign="top">

Screenshot 2022-04-06 at 13 43 45

 </td>
<td valign="top">

Screenshot 2022-04-06 at 10 35 52

</td>
</tr>
</table>

#### TypeScript format
<table>
<tr>
<th> Before </th> <th> After </th>
</tr>
<tr>
<td valign="top">

Screenshot 2022-04-06 at 10 36 51

 </td>
<td valign="top">

Screenshot 2022-04-06 at 10 35 36

</td>
</tr>
</table>

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -59,8 +59,11 @@ jobs:
       - name: Install dependencies
         run: yarn
 
-      - name: Build
+      - name: Build tokens
         run: yarn build
+
+      - name: Build experimental tokens
+        run: yarn build:tokens
 
       - name: Install dependencies (base)
         run: pushd base; yarn; popd
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,12 @@ jobs:
       - name: Install dependencies
         run: yarn
 
+      - name: Build tokens
+        run: yarn build
+
+      - name: Build experimental tokens
+        run: yarn build:tokens
+
       - name: Create release pull request or publish to npm
         id: changesets
         uses: changesets/action@master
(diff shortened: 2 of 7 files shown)


langermank (commented):
Tested the build, looks good! Feel free to merge this into my branch.

If you also want to add that separate build script we discussed, go for it 👍

rezrah:
> If you also want to add that separate build script we discussed, go for it 👍

Done ✅ . CI workflows now use both build stages while the local developer experience is unaffected and only builds the main tokens. This should allow the base branch to safely merge into `main` if you want @langermank.

PR #318, opened 2022-04-20
add deployment workflows for GitHub Pages

rezrah:
### What are you trying to accomplish?

Part of https://github.com/github/primer/issues/886

Replaces deployment target for all docs from Vercel to GitHub Pages

### What approach did you choose and why?

Utilises reusable workflows make future deployments consistent with other Primer repo's.

### What should reviewers focus on?

- Use the new preview urls, which will appear in all new PRs - including this one - as follows:
   
  Screenshot 2022-04-19 at 14 41 00
  ☝️  Look for this comment at the bottom of the PR

  or

  Screenshot 2022-04-19 at 14 41 10
  ☝️  find the latest deployment preview URL in the Environments section at this bottom of this PR

- Review the latest preview deployment and check everything is appearing correctly 🙏 

### Todo

- [x] Create environments
- [x] Configure feature flags

### Post-merge 

- [ ] Remove Vercel environment
- [ ] Remove project and deployment triggers in Primer Vercel account

PR #321, opened 2022-04-20
Removes postinstall command

rezrah:
Adding `postinstall` in #318  causes errors the released package. This fixes it by removing postinstall and triggering a new patch release.


langermank (approved):
Tested the pre-release, seems to work!

PR #323, opened 2022-04-25
add self serve tooling for new primitives

rezrah:
### What's changing?

Refactoring v2 tokens to build and export self-serve primitives from `@primer/primitives`

**🎥  Check out the demo:** https://www.loom.com/share/a300e866498844bba551fc4b5c6795ae
 
### Why?

Part of https://github.com/github/primer/issues/924

Primer Brand is expected to generate its own primitives, rather than importing the product tokens from in `@primer/primitives`. 

See epic for more details: https://github.com/github/primer/issues/924

### How?

- Converts previous build functions to usage-agnostic tooling
  - Internally dogfoods the tooling to continue building product-specific tokens
  - Exports the tooling so that Primer Brand can build the same tokens, but also override them where necessary

### DIFF
--- a/.changeset/pink-rice-tell.md
+++ b/.changeset/pink-rice-tell.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+add self serve tooling for v2 primitives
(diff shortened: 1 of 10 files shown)


langermank (approved):
Incredible work Reza! I pulled and built locally, everything looks good to me 👍

PR #324, opened 2022-04-26
[WIP] Support 3:1 contrast minimum for interactive controls

mperrotti:
**_Implementation is on pause until we're aligned with a11y, brand, and design infra teams._**

Related to https://github.com/github/primer/issues/836

The deployments from this PR are being used to test style updates in Primer CSS and Primer React.

`@primer/css` update PR: https://github.com/primer/css/pull/2045
dotcom update branch: [`mp/control-min-contrast`](https://github.com/github/github/pull/new/mp/control-min-contrast)

### DIFF
--- a/data/colors/vars/component_light.ts
+++ b/data/colors/vars/component_light.ts
@@ -187,5 +187,9 @@ export default {
       bg: get('scale.blue.5'),
       disabledBg: get('scale.gray.5'),
     }
+  },
+
+  control: {
+    border: get('scale.gray.3') // maintains min 3:1 contrast ratio
   }
 }
--- a/data/colors/vars/component_light.ts
+++ b/data/colors/vars/component_light.ts
@@ -190,6 +190,6 @@ export default {
   },
 
   control: {
-    border: get('scale.gray.3') // maintains min 3:1 contrast ratio
+    border: get('scale.gray.4') // maintains min 3:1 contrast ratio
   }
 }
--- a/.changeset/cool-drinks-occur.md
+++ b/.changeset/cool-drinks-occur.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
+Add border with 3:1 contrast ratio for interactive controls
--- a/data/colors/themes/dark_high_contrast.ts
+++ b/data/colors/themes/dark_high_contrast.ts
@@ -223,6 +223,10 @@ const exceptions = {
       border: get('scale.blue.5'),
     }
   },
+
+  control: {
+    border: get('border.default') // maintains min 3:1 contrast ratio
+  }
 }
 
 export default merge(dark, exceptions, {scale})
--- a/data/colors/themes/light_high_contrast.ts
+++ b/data/colors/themes/light_high_contrast.ts
@@ -235,6 +235,9 @@ const exceptions = {
       border: get('scale.blue.5'),
     }
   },
+  control: {
+    border: get('border.default') // maintains min 3:1 contrast ratio
+  }
 }
 
 export default merge(light, exceptions, {scale})
--- a/data/colors/vars/component_dark.ts
+++ b/data/colors/vars/component_dark.ts
@@ -186,5 +186,9 @@ export default {
       bg: get('scale.blue.5'),
       disabledBg: get('scale.gray.5'),
     }
+  },
+
+  control: {
+    border: get('scale.gray.4') // maintains min 3:1 contrast ratio
   }
 }


mperrotti:
@langermank - yup! I'm going to post screenshots as soon as I make changes in Primer CSS and Primer React.

simurai (commented):
Do we already know what components will use this new `control.border`? I remember @vdepizzol mentioning that if a control already has text or an icon with enough contrast it's ok to keep using the current `border.default`. I think what's described [here](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#:~:text=This%20success%20criterion%20does%20not%20require%20that%20controls%20have%20a%20visual%20boundary%20indicating%20the%20hit%20area%2C%20but%20if%20the%20visual%20indicator%20of%20the%20control%20is%20the%20only%20way%20to%20identify%20the%20control%2C%20then%20that%20indicator%20must%20have%20sufficient%20contrast.).

 So is it mostly for `form-control` where there is a chance that the input is empty without placeholder text:
  
image

If so, I wonder if this token should explicitly mention that it's meant for creating enough contrast e.g. `control.border.accessible`, `control.border.contrast` or `control.border.boundary`.

Another thought... since `control` isn't an existing component and more a shared token, should it be added to the [`primer.border`s](https://github.com/primer/primitives/blob/94d58003c1509d2a79ee9639f2b657cdc549e9de/data/colors/vars/global_light.ts#L99-L102). Just for now until we move all colors to the new [tokens](https://github.com/primer/primitives/tree/main/tokens).

mperrotti:
Thanks for taking a look @simurai !

> Do we already know what components will use this new control.border?

I did an audit yesterday, which you can read more about in [this comment](https://github.com/github/primer/issues/836#issuecomment-1110272361).

So far, it looks like we only need this for a text inputs. I'm also going to update the `<select>` border so it matches the text input. We use user agent styles for checkboxes and radios, and those meet the 3:1 minimum.

---

> I remember @vdepizzol mentioning that if a control already has text or an icon with enough contrast it's ok to keep using the current border.default.

Correct - if a control has text or an icon with enough contrast, it's ok to keep using the current `border.default`.

---

> Another thought... since control isn't an existing component and more a shared token, should it be added to the [primer.borders](https://github.com/primer/primitives/blob/94d58003c1509d2a79ee9639f2b657cdc549e9de/data/colors/vars/global_light.ts#L99-L102). Just for now until we move all colors to the new [tokens](https://github.com/primer/primitives/tree/main/tokens).

I considered adding this to the [primer.borders](https://github.com/primer/primitives/blob/94d58003c1509d2a79ee9639f2b657cdc549e9de/data/colors/vars/global_light.ts#L99-L102) as `primer.borders.control`, but I wasn't sure if it was ok to break from the other colors that just have `emphasis`, `muted`, `subtle`.

Now that I'm looking at how we treat `canvas` and `shadow` differently, I'm thinking it's safe to have `primer.borders.control`. I'll update my this PR with that change.

vdepizzol:
> Now that I'm looking at how we treat canvas and shadow differently, I'm thinking it's safe to have `primer.borders.control`. I'll update my this PR with that change.

I'm a little concerned about this name since not all control borders will need a 3:1 contrast requirement, and `borders.control` is not specific enough.

I'm starting to put together [control-related design tokens for colors](https://www.figma.com/file/CN55YCXQYi7vec8CHZqy2X/Design-tokens-(FY22-Q4)?node-id=2%3A2316) (only available for Hubbers), and SegmentedControl will also benefit from the 3:1 contrast for selected items.

It would be nice to have a system-level thinking that takes into consideration the new [design token naming guidelines](https://github.com/github/primer/blob/main/adrs/2022-03-31-design-token-naming-guidelines.md), so we don't need to update this again in the future. 🙇

simurai:
> I'm a little concerned about this name since not all control borders will need a 3:1 contrast requirement, and `borders.control` is not specific enough.

How about `primer.control.border.emphasis`?

- `primer` (namespace)
- `control` (item)
- `border` (property)
- `emphasis` (modifier)

A control border that has extra **emphasis**.

langermank:
> How about primer.control.border.emphasis?

I was thinking something along these lines as well! Our modifier should be able to be extended for multiple variations. Since this darker shade will be the default, I wonder if `emphasis` would be clear enough since it almost reads like "accent". I took a closer look at a few control and field type components (some new and some existing) and I feel like we may be able to support most (maybe all?) cases by providing two variables:

`primer-control-border-mid`: `color-border-default` aka `scale-gray-2`
`primer-control-border-dark`:  `scale-gray-4` slightly lightened to meet 3:1 minimum

I think down the line we could consider this naming for base borders and remove "default" from the vocabulary:

```
base-border-lightest
base-border-light
base-border-mid
base-border-dark
```

If we want to be more specific, we could consider: 
`primer-control-border-muted`:  to match a "muted" style prop for inputs that have leading visuals
`primer-control-border-regular`: to signify this is the default

form control prop options

form controls on a gray background with different contrast borders

data/colors/vars/component_light.ts
```diff
@@ -187,5 +187,9 @@ export default {
       bg: get('scale.blue.5'),
       disabledBg: get('scale.gray.5'),
     }
+  },
+
+  control: {
+    border: get('scale.gray.4') // maintains min 3:1 contrast ratio
```
langermank, line 193: I've only tested this in `light` but we should be able to slightly lighten this and still meet the minimum requirement: ``lighten(get('scale.gray.4'), -0.03)``

simurai:
> form control prop options

👍  Using a regular border when there is an icon would also allow us to keep these input/button combos like in the [Account settings](https://github.com/settings/admin):

image

Although one concern is that it might feel confusing when two inputs are close to each other but use two different borders. You could mistake one input is already focused or the other is disabled? Couldn't find an example, so maybe not a big issue, but something like this mockup:

Screen Shot 2022-05-04 at 12 52 40

mperrotti:
Thank you everybody for all of the thoughtful feedback! There are so many good ideas in here that I think we could bring to our border primitives.

However, when I was testing these new borders in dotcom, the darker input borders had a pretty negative impact on our overall aesthetic. I brought this up during the DI design sync ([see notes](https://docs.google.com/document/d/1nVJoI23GT-NNYzIIXJmvXGIxAniUjJX-a1WBJCVpM7c/edit#heading=h.26ub8cr0xo9z) - internal folks only), and we may be able to avoid adding this altogether by relying on high-contrast mode.

langermank:
Closing this as we have a new token build system in place. I think there's more work to do with this in the future which we can address in a new PR.

PR #334, opened 2022-06-22
Typography bug fixes + border thickness naming

langermank:
- Adjust border thickness naming
- Use unitless `calc()` for `line-height`

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -71,6 +71,9 @@ jobs:
       - name: Build (base)
         run: pushd base; yarn build; popd
 
+      - name: Build (base experimental)
+        run: pushd base; yarn build:tokens; popd
+
       - name: Diff
         uses: primer/comment-token-update@main
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/diff.yml
```diff
@@ -84,7 +87,7 @@ jobs:
           # this action may not work as expected.
           comment-token: 'diff'
           script: |
-            diff=$(for file in dist/scss/**/*.scss
+            diff=$(for file in dist/scss/**/*.scss, tokens-v2-private/css/tokens/**/*.css
```
langermank, line 90: @rezrah I decided to mess about with this to see if I could get diff comments to work for the newbies. One of my commits I was able to get it to work 😆 but only if I specified the literal filepath to the file I knew changed. Any thoughts?

langermank:
Diff that I hacked together 😄 

css variable changes for line-height

.github/workflows/diff.yml
langermank: We decided to worry about this next quarter 😄

rezrah (approved):

PR #336, opened 2022-06-30
add system-ui to base fonts

keithamus:
This adds `system-ui` to our base font tokens.

This value represents the same font that the operating system can use. On Safari `system-ui` is the same as `-apple-system`, and on Chrome on Mac it is the same as `BlinkMacSystemFont`.

CSS spec: https://drafts.csswg.org/css-fonts-4/#valdef-font-family-system-ui
Chrome Status: https://chromestatus.com/feature/5640395337760768 - implemented in Chrome 56+, Edge 79+
Webkit Bug: https://bugs.webkit.org/show_bug.cgi?id=151493 - implemented in Safari 11+
Firefox Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1226042 - implemented in Firefox 92.

### DIFF
--- a/tokens/functional/typography/typography.json
+++ b/tokens/functional/typography/typography.json
@@ -2,10 +2,10 @@
   "<namespace>": {
     "fontStack": {
       "system": {
-        "value": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'"
+        "value": "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'"
       },
       "sansSerif": {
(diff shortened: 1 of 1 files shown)


simurai (changes requested):
We tried this before, but reverted it again due to issues in Chinese https://github.com/primer/css/issues/1589. So we would need to test first if it's still causing problems.

Another option that I think came up in Slack that I quite like:

- Use `system-ui` as the default
- But let users change the font in the user settings (for cases where the system font causes issues e.g. Chinese)

Similar like code editors let you pick your preferred font:

image

lukasoppermann:
@keithamus @simurai can we close this? Does it not work? Or can we somehow continue to merge this at some point?

simurai:
I think we could try `system-ui` again at _some point_, but not sure if it's still too early. Like there are millions of users and some might still use an outdated OS/browser and as mentioned we had to revert it due to problems with [Chinese](https://github.com/primer/css/issues/1589).

Also some time later we merged https://github.com/primer/primitives/pull/385 that should improve fonts for Linux and even with that, we got some backlash on Twitter because some didn't like the change.

So personally, I would still wait a bit to enable `system-ui`.

PR #338, opened 2022-07-10
Normalize color data to lowercase hex

adrian5:
Doesn't matter as much in code, but kinda stood out to me on the Primer website.

P.s. is there a reason some color arrays are in a single line and others in a vertical list, or is that random (different contributors)?


adrian5:
Understood, thanks!

PR #345, opened 2022-08-31
Started to added jest testing

lukasoppermann:
As discussed with @langermank it would be great to add a test suit for the token transformation. This is the beginning.

## Why?
Primitives are a central element of primer. Breaking them or the token pipeline would result in potential issues in the product.
To avoid this and still allow us to evolve the system we need a baseline to tell us when we change behavior.
This is why having tests is essential.

## How
My idea is to start with jest and write tests for the style dictionary transformation. 
This is in addition to the diffing that @rezrah will add, so that we see when the output changes.

This should allow us to iterate on the tokens and the transformation while keeping an eye out for any undesired change.

## Big picture
Ideally we want to transform the repo to make it easier for people to understand, reason about and contribute.
This also means things like splitting up the build file into smaller packages that abstract functionality away.

Our hope is to have a healthy codebase and encourage people to contribute.


package.json
```diff
@@ -32,7 +32,8 @@
     "prebuild:tokens": "rm -rf tokens-v2-private",
     "prepack": "yarn build && yarn build:tokens",
     "release": "changeset publish",
-    "watch": "ls data/**/*.scss script/**/*.ts | entr -s 'yarn build'"
+    "watch": "ls data/**/*.scss script/**/*.ts | entr -s 'yarn build'",
+    "test": "jest"
```
rezrah, line 36: To preserve ordering..

```suggestion
    "test": "jest"
    "watch": "ls data/**/*.scss script/**/*.ts | entr -s 'yarn build'",
```

package.json
```diff
@@ -57,11 +59,13 @@
     "deep-shape-equals": "^0.1.2",
     "deepmerge": "^4.2.2",
     "flat": "^5.0.2",
+    "jest": "^29.0.1",
     "lodash": "^4.17.20",
     "mkdirp": "^1.0.4",
     "node-sass": "^7.0.0",
     "sass-extract": "^2.1.0",
     "style-dictionary": "^3.1.1",
+    "ts-jest": "^28.0.8",
```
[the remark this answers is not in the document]
rezrah: Seems to be have a peer dependency issue too on a fresh install...
Screenshot 2022-09-06 at 10 59 26

package.json
rezrah, line 36: Totally optional, but I like running this with the following flags `--verbose --coverage` because it gives additional context into what ran and how tests performed..

```suggestion
    "test": "jest --verbose --coverage"
```

package.json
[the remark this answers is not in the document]
lukasoppermann: Nope, we can move to normal jest

__tests__/pathToDotNotation.test.js
```diff
@@ -0,0 +1,13 @@
+const pathToDotNotation = require('../build.js').pathToDotNotation
+
+test('valid token', () => {
```
rezrah, line 3: Could we wrap this in a describe block? It will help to provide that additional context if we add verbose logging and coverage reports...

```suggestion
describe('Token formatters: PascalCase', () => {
  it('renders a valid token name in PascalCase', () => {
```

__tests__/pathToKebabCase.test.js
```diff
@@ -0,0 +1,13 @@
+const pathToKebabCase = require('../build.js').pathToKebabCase
+
+test('valid token', () => {
```
rezrah, line 3:
```suggestion
it('returns a valid token name in kebab-case', () => {
```

also recommend wrapping these in a describe block too ideally...

__tests__/pathToDotNotation.test.js
```diff
@@ -0,0 +1,14 @@
+const pathToDotNotation = require('../build.js').pathToDotNotation
+
+describe('Token formatters: PascalCase', () => {
+  it('returns a valid token name in PascalCase', () => {
```
rezrah: oof, my bad.. my suggestion earlier was for the wrong file...

this is the correction for this file..

```suggestion
describe('Token formatters: dot notation', () => {
  it('returns a valid token name in dot notation', () => {
```

sorry about that @lukasoppermann 😅

build.js
```diff
@@ -3,6 +3,17 @@ const StyleDictionary = require('style-dictionary')
...
+//-----
+// functions to be extracted
+// TODO: extract to a separate files
+
+const pathToKebabCase = token => token.path.join('-')
+
+const pathToDotNotation = token => token.path.join('.')
+
+const capitalize = string => string[0].toUpperCase() + string.slice(1)
+const pathToPascalCase = token => token.path.map(tokenPathItems => capitalize(tokenPathItems)).join('')
```
rezrah, line 15: being pedantic, can we add a newline here like the others please? 😅 

```suggestion

const pathToPascalCase = token => token.path.map(tokenPathItems => capitalize(tokenPathItems)).join('')
```

__tests__/pathToDotNotation.test.js
lukasoppermann: I think I fixed it already when I added the blocks.

rezrah:
@lukasoppermann - FYI I've pushed a fix for that compiler error by upgrading to v4 TypeScript. Adding Jest had bumped the nested devDependency for TS that caused the issue.  Builds seem to be passing now.

rezrah (approved):
Thanks for doing this @lukasoppermann. It works great! 

Tested end-to-end in Primer Brand. No regressions 🎉.

PR #347, opened 2022-09-07
PoC of extracting functions and bundling ts file

lukasoppermann:
Hey @rezrah I build a first PoC for extracting functions and bundling to js afterwards,

### What this PR adds
- extract naming fns into `nameTransformer.ts`
- change `build.js` to `build.ts`
- add minimal Token type

### Limitations
- to avoid adding `ts-jest` for now, the tests are using the bundled `build.js` file from dist, this should be changed by fixing the dependency issue upstream
- I created a new tsconfig to specifically bundle this build file
- typescript errors need to be fixed in `build.ts`
- vs code complained about the `Token` type if I don't import it, is there a workaround as it is within the dir that is defined in `tsconfig.json`


newToken.tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
```
rezrah, line 2: There appears to already be a tsconfig in the project. Do we need another one? Or is this doing something the other isn't?

package.json
```diff
@@ -27,12 +27,14 @@
   "homepage": "https://github.com/primer/primitives#readme",
   "scripts": {
     "build": "ts-node ./script/build.ts && tsc",
-    "build:tokens": "node -e \"require('./build')._init()\"",
+    "build:tokens": "ts-node -e \"require('./build')._init()\"",
+    "build:buildJs": "tsc --project newToken.tsconfig.json",
```
rezrah, line 31: I think we could run tsc and ts-node sequentially to avoid having two commands.

src/@types/token.d.ts
```diff
@@ -0,0 +1,3 @@
+export type Token = {
```
rezrah, line 1: StyleDictionary exports some type definitions, could we pull from there instead? Also, this folder is called `@types`. Is there a relation to the `@types` package? I'd recommend we move this into `nameTransformer.ts` for now and let `tsc` create the `d.ts` during compile time.

newToken.tsconfig.json
lukasoppermann: The normal tsconfig is sued for the `build.ts` in `scripts/` which is not the same. The goal is to combine all, but for this we need to port colors.

Ideally we would only have one, but I tried to get this bundling start with the least changes possible.

package.json
lukasoppermann: well, we don't need `tsc` when we build the tokens. And once we get `ts-jest` we only need it for `primer/brand` this is why I think we could keep it separate.

src/@types/token.d.ts
lukasoppermann: Good one, I'll change it, it is `TransformedToken`

newToken.tsconfig.json
rezrah: okay, I think you could just extend from that original version then? and add only the delta between the two to this new one? 

```json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    ...new stuff only for your build file
  }
}
```

lukasoppermann: Yes, we can do that but most props are overwritten. When colors are moved we can hopefully remove the old one in any case

rezrah: 🤔 okay, to avoid any confusion for other contributors in the interim, can you rename the file to be `v2-tokens.tsconfig.json` or similar instead of `new.tsconfig.json`?

package.json
rezrah: cool cool 👍 ... how about `build:self-serve` then? `build:buildJs` could be too ambiguous for new folks browsing the repo.

lukasoppermann:
All done in individual PRs.

PR #349, opened 2022-09-08
color contrast check prototype

lukasoppermann:
This is just the start. Many questions are popping up:

1. where do we store the contrast requirement (directly in `token.json`? In a separate spec file?)
2. Should a build fail if contrasts fail? (Currently impossible, so we would need to add an `ignore` flag that will report the fail but ignore it)
3. Should we ignore component tokens and run it on primitives only (I think yes)
4. We need to import the json when we have one for the new build process and use that to test

## Todo:
- [ ] replace lib with own code to reduce dependency (if possible) 
- [ ] add better table (left aligned, not `'` around strings)
- [ ] clean up code to make it easier to read
- [ ] add all pairs
- [ ] make it run on all modes (rules may differ depending on mode, e.g. HC)

### DIFF
--- a/script/color-contrast.ts
+++ b/script/color-contrast.ts
@@ -1,0 +1,53 @@
+// @ts-ignore
+import getContrastRatio from 'get-contrast-ratio';
+import colors from "../dist/ts"
+
+const flattenObject = (obj: any, prefix: string = "") =>
+  // @ts-ignore
+  Object.keys(obj).reduce((acc, k) => {
+    const pre = prefix.length ? prefix + '.' : '';
+    if (typeof obj[k] === 'object') Object.assign(acc, flattenObject(obj[k], pre + k));
+    // @ts-ignore
+    else acc[pre + k] = obj[k];
+    return acc;
+  }, {})
+
+// const runContrastTest = ([required, ...pair]: (number | string)[], colors: any) => {
+const runContrastTest = (colorPairs: (number | string)[][], colors: any) =>
+  Object.fromEntries(
+    colorPairs.map(([required, ...pair]: (number | string)[]) => {
+      // concat name
+      const contrastPair = pair.join(' vs. ')
+      // @ts-ignore
+      const [pass, contrast] = testContrast(required, colors[pair[0]], colors[pair[1]])
+      // build required string
+      const requiredContrast = `${required}:1`
+      // 
+      return [
+        [contrastPair], {
+          pass,
+          contrast,
+          requiredContrast
+        }
+      ]
+    })
+  )
+
+const testContrast = (required: number, colorA: string, colorB: string): [pass: string, contrast: string] => {
+  // get contrast
+  const contrast = getContrastRatio(colorA, colorB)
+  return [
+    contrast >= required ? '✅' : '❌',
+    `${contrast}:1`
+  ]
+}
+
+const colorPairs = [
+  [4.5, 'fg.default', 'canvas.default'],
+  [4.5, 'fg.muted', 'canvas.default'],
+  [4.5, 'fg.default', 'canvas.subtle'],
+  [4.5, 'fg.muted', 'canvas.subtle']
+]
+
+const flattenColors = flattenObject(colors.colors.light)
+console.table(runContrastTest(colorPairs, flattenColors))
--- a/package.json
+++ b/package.json
@@ -2,23 +2,11 @@
   "name": "@primer/primitives",
   "version": "7.9.0",
   "description": "Typography, spacing, and color primitives for Primer design system",
-  "files": [
-    "dist",
-    "tokens-v2-private",
-    "build.js",
-    "tokens/"
-  ],
+  "files": ["dist", "tokens-v2-private", "build.js", "tokens/"],
   "main": "dist/js/index.js",
   "types": "dist/js/index.d.ts",
   "repository": "https://github.com/primer/primitives",
-  "keywords": [
-    "primitives",
-    "colors",
-    "spacing",
-    "typography",
-    "variables",
-    "design-system"
-  ],
+  "keywords": ["primitives", "colors", "spacing", "typography", "variables", "design-system"],
   "author": "GitHub, Inc.",
   "license": "MIT",
   "bugs": {
@@ -30,6 +18,7 @@
     "build:tokens": "node -e \"require('./build')._init()\"",
     "prebuild": "rm -rf dist",
     "prebuild:tokens": "rm -rf tokens-v2-private",
+    "postbuild": "ts-node ./script/color-contrast.ts",
     "prepack": "yarn build && yarn build:tokens",
     "release": "changeset publish",
     "test": "jest --verbose --coverage",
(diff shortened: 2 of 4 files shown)


lukasoppermann:
I added all color pairs now. However the lib does not work with transparent colors.

This is something we need to address. The "problem" here is that we may transparent colors on multiple bgs.

E.g. a `accent.muted` could be place on `canvas.default` or `canvas.inset`. So we would probably need to add to some pairs the possibilities of background colors.

This should work but adds more effort.

@rezrah do you have any other idea?

lukasoppermann:
Okay, I added the possibility to define multiple bgs for transparent colors.

The way it works now:

- for each pair you can define an options object with a `bgs: string[]` property, the bg strings are references to colors, e.g. `canvas default`.
- the script creates a check for each specified bg
- when testing you have colorA, colorB, bg
  - first colorB is mixed with bg to create an opaque version of colorB on bg
  - colorA is mixed with colorB to create an opaque color
  - both colors are tested against each other

lukasoppermann:
Hey @rezrah I think this is ready now.
If you have an idea how to replace any of the dependencies feel free to let me know.

Also this script outputs `color-contrast-check.json`. This can be used for the workflow part, but I think the workflow is step 2, correct? So we merge this first and do the workflow afterwards.

script/color-contrast.config.ts
```diff
@@ -0,0 +1,225 @@
...
+  light_high_contrast: highContrast,
+  light_colorblind: baseRequirements,
+  light_tritanopia: baseRequirements,
+  // default dark mode
+  dark: baseRequirements,
+  dark_dimmed: baseRequirements,
+  dark_high_contrast: highContrast,
+  dark_colorblind: baseRequirements,
+  dark_tritanopia: baseRequirements,
+}
```
rezrah, line 225: Question around maintainability of this file in future..

Is it worth co-locating this information to the the functional src tokens as additional metadata per token? If the plan is to add _more_ functional tokens in future to Primitives, it can be easy to forget to update this file. We could also do something interesting by using style dictionary custom JSON format  to generate the accessibility lookup tables.

Haven't thought this through entirely yet, but the concern is coming from risks to long-term maintainability of this file. That said, if I'm wrong on my assumption that we add functional tokens regularly, this is a moot point.

script/color-contrast.ts
```diff
@@ -0,0 +1,173 @@
+// @ts-ignore
```
rezrah, line 1: What was the error here?

script/color-contrast.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+/**
+ * renderConsoleTable
+ * @description takes the test results per theme and prints a nicely formatted table of the results to the console 
+ * @param theme 
+ * @param results 
+ */
+const renderConsoleTable = (theme: string, results: contrastTestResult[]): void => {
+  // config table
+  const contrastTable = new Table({
```
rezrah, line 107: Any chance we can do this without the dependency? Appreciate it's the quicker option, but it leaves us open to security and maintenance burden over time that becomes difficult to justify for otherwise straightfoward tasks. Happy to help rewrite this in a part 2 PR if that helps too!

script/color-contrast.ts
```diff
@@ -0,0 +1,173 @@
...
+    renderConsoleTable(theme, results)
+    // return results for json file creation
+    return {
+      theme,
+      results
+    }
+  })
+
+// write json file for workflow
+fs.writeFile('color-contrast-check.json', JSON.stringify(results), (err) => {
```
rezrah, line 169: Can you put the results into the `dist/` folder please? That's on a gitignore so it won't create any unstaged changes on the CI server when/if we do auto-commits from there. I'll still be able to use it for workflows from there. 

```suggestion
const outputDirectory = '/tokens-v2-private/json';
if (!fs.existsSync(outputDirectory)){
    fs.mkdirSync(outputDirectory);
}
fs.writeFile(`${outputDirectory}/color-contrast-check.json`, JSON.stringify(results), (err) => {
```

rezrah (commented):
This is great @lukasoppermann 🙌 

- To answer your question in description, I can definitely work with this in CI workflows
- Left some comments on things we should fix before this merges. Typing errors, deps etc.
  - Happy to help fix those with you or in future PR if easier too. LMK.

> Should a build fail if contrasts fail?

IMO no, because this is a side-effect and additional abstraction on top of build jobs. It would be inaccurate to suggest the build failed when it was actually a governance layer that did. We can work around it by making the contrast check fail instead for transparency.

Instead, let's focus on failing package releases:

- Canary should always pass and ship, to avoid blocking testing
- GA on `@latest` channel should fail

Thoughts?

script/color-contrast.ts
```diff
@@ -0,0 +1,173 @@
...
+      minimumContrastRatio
+    })
+    )
+  })
+// )
+/**
+ * testContrast
+ * @description test the contrast of two colors against each other
+ * @param minimumContrast 
+ * @param colorA
```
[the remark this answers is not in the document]
lukasoppermann: We could, but this would not be truthful for testing e.g. default text color vs. link color. I had fg vs. bg first.

Happy to adjust it again, what do you think?

script/color-contrast.ts
lukasoppermann: Yeah, lets book this for a 2nd PR please.

Reasons for the dep:
- Table looks better
- center aligned columns with `console.table` are hard to read

script/color-contrast.config.ts
lukasoppermann: I would love this. Adding a metadata part on the tokens sounds pretty good. Than we can do something like with `deprecated`. It could also be used for docs to show contrast ratios, etc. if we want. And to suggest pairing in `vscode` if this is possible.

lukasoppermann: But this would be after the PR is merged. I'll create an issue for it.

script/color-contrast.ts
lukasoppermann: Gone, don't know what it was.

script/lib/flattenObject.ts
```diff
@@ -0,0 +1,9 @@
+export const flattenObject = (obj: any, prefix: string = "") =>
```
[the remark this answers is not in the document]
lukasoppermann: Actually lodash does not seem to have flatten for objects where the name is joined with a dot. Left it in but moved it.

script/color-contrast.ts
lukasoppermann: Done.

script/color-contrast.ts
rezrah: >Table looks better 

FYI.. worth bearing in mind that comments in the Pull Request will strip out all extra styling

script/utilities/flattenObject.ts
```diff
@@ -0,0 +1,9 @@
+export const flattenObject = (obj: any, prefix: string = "") =>
+  // @ts-ignore
```
rezrah, line 2: Can this one be removed too?

script/color-contrast.ts
lukasoppermann: But not when using it locally.

script/utilities/flattenObject.ts
lukasoppermann: Yes

PR #351, opened 2022-09-14
Add SCSS vars for PCSS utilities & dotcom

langermank:
This PR adds a format for SCSS variables. While we aim to move towards CSS variables, SCSS is still deeply embedded in PCSS and dotcom. I'd like to continue offering SCSS vars with our newer set of values (rem units) as a stopgap before we have wider adoption of CSS variables.

- SCSS vars can support some of the legacy email CSS
- Math without `calc()` will continue to work if we choose to swap in new values for existing PCSS variables
- We may consider using SCSS vars for utility classes so they can remain immutable

### DIFF
--- a/build.js
+++ b/build.js
@@ -18,6 +18,16 @@ const pathToPascalCase = token => token.path.map(tokenPathItems => capitalize(to
 // REGISTER THE CUSTOM TRANFORMS
 
 /**
+ * transform: scss variable names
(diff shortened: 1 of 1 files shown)


rezrah (approved):
Good idea to incrementally work towards the css vars. Changes LGTM too. 👍 

Sidenote: lots of formatting changes here. Any issues if we put Prettier in this repo?

PR #353, opened 2022-09-19
Style dictionary update including adding colors

lukasoppermann:
As mentioned to @rezrah in the loom, I am trying to get our config back to the SD basics and at the same time extract small functions.

Additionally I want the json files to be the only influence on the token output so that a user can just edit json files and does not have to expect any side effects form the script.

**Focus:** easy authoring & json that conform to [w3c design token spec](https://design-tokens.github.io/community-group/format/)

## Todos
- [x] **Add custom alpha function:** https://github.com/lukasoppermann/style-dictionary-utils/blob/main/src/utilities/alpha.ts because the color2k _transparentize_ function has very inconsistent behavior and does not work as we work with alpha
- [ ] Platforms:
  - [x] CSS
  - [x] SCSS with css variables
  - [x] JS as commonJs module
  - [x] TS as es6 `export`
  - [ ] Typescript definitions
  - [x] JSON for docs
  - [ ] JSON for figma
  - [x] JSON for vscode → **no changes in output**
  - [x] Deprecated JSON → **no changes in output**
    - [x] individual files for `color.json`, `size.json` , `typograhy.json` 
    - [x] build deprecated.json from tokens with `deprecated` property like @rezrah mentioned here: https://github.com/github/primer/pull/1309#discussion_r978409181
  - [x] Removed JSON → File is just copied over → **no changes in output**
- [x] `JSON5` parser: https://github.com/primer/primitives/issues/357
- [x] Color tokens
  - [x] create scale source tokens
  - [x] ~don't export base~ pipe base scales to output: https://github.com/github/github/pull/240557
  - [x] create function source tokens
  - [x] scales & merge with colorblindness scales
  - [x] exceptions for functional tokens per color mode
- [ ] **Shadows:**
  - [ ] not correct as is not referencing color but moved out of color modes
  - [ ] may be solved by #359 → make shadow transform color with alpha values so we don't need specific shadow colors
  - [ ] if not solved by #359 we need to build shadows with reference to shadowColor
- [ ] Figma json
  - [x] filter out all tokens that are not `$type=color`, `$type=typography` and have no `figmaType`
  - [ ] add `figmaType` to all tokens where applicable according to [figmaTokens](https://docs.figmatokens.com/tokens/available-tokens)
- [x] [FontWeight](https://design-tokens.github.io/community-group/format/#font-weight) transformer to number
- [x] [Dimension](https://design-tokens.github.io/community-group/format/#dimension) transformer (to px or to rem)
- [x] [Shadow](https://design-tokens.github.io/community-group/format/#shadow) transformer
- [x] Transform colors alpha value when `alpha` is present
  - [x] What happens if `value` is a color that is already transparent? → Alpha is added up (0.6 - 0.4 => 0.2)
- [x] Update source token files to be w3c conform
  - [x] add `$type`
  - [x] change `value` to `$value`
  - [x] ~add required `letterSpacing` to `typography` tokens~ [hoping for w3c to change typography to not require all values](https://github.com/design-tokens/community-group/issues/102#issuecomment-1261887142)
- [x] add [functional font family tokens
](https://github.slack.com/archives/C03UZDAUYF8/p1664468300304749?thread_ts=1664439730.893449&cid=C03UZDAUYF8) `ui`, `code`, `content` and move `system, sansSerif, serif, monospace` to `base`

## Cleanup
Could also be tackled after merging.
- [ ] Cleanup build file and how different builds are started
- [ ] Is there a better way to import color json files (currently it may be confusing for new people) at least storing names in short variables could improve readability
- [ ] Organize token folder?
  - [ ] base color tokens should files be prefixed with `extension` to show that they extend `dark.json` or `light.json`?
  - [ ] functional should `exception` files be placed in a subfolder?
  - [ ] should we split between dark and light mode into subfolders?

## Questions
- [ ] How do we deal with `prefix`: https://github.com/primer/primitives/pull/353#issuecomment-1252207814


lukasoppermann:
Hey @langermank & @rezrah, I am currently thinking about the prefix.

When do we want to have the `primer` prefix and for which outputs?

You are export with and without here: https://primer.style/primitives/typography and for sizes.
But only base has no prefix.

It also seems we don't have prefixes here: https://primer.style/react/theming or here https://unpkg.com/@primer/css@20.4.5/dist/primer.css

So should we not add prefixes?

Lastly I am not sure adding prefixes to js objects makes sense as you can prefix them when importing and with prefix you would get something like `theme.primer.color.canvas.default`.

```js
import theme from '@primer/ts/light.ts'

lukasoppermann:
@langermank another complication I ran into is: If we don't want to keep references to the base scale in the code (e.g. `var(--primer-base-size-12)`) but we do want to keep references to other functions files (e.g. `--primer-viewport-range-wide: (min-width: var(--primer-breakpoint-xxlarge));`) it will require a lot of working around this.

Is that really necessary?

langermank:
Re: prefix- the prefix behavior we've implemented is part of what ties primer/primitives together with primer/brand. The default is `primer` which is used for all product tokens, and in brand we use `brand`. @rezrah can go into detail about how the build process works, but this allows brand to replace tokens with their own prefix. If a token value is redefined in brand, it will simply replace the primer/primitive instead of sitting alongside it with a different prefix.

re: base scale- the only base scale we don't want to keep references for are base scale colors. Everything else is fair game.

lukasoppermann:
> Re: prefix- the prefix behavior we've implemented is part of what ties primer/primitives together with primer/brand. The default is primer which is used for all product tokens, and in brand we use brand. @rezrah can go into detail about how the build process works, but this allows brand to replace tokens with their own prefix. If a token value is redefined in brand, it will simply replace the primer/primitive instead of sitting alongside it with a different prefix.

I think you misunderstood this part. But we probably want to keep the output as close the the current output as we can, right? So I will not be adding a `prefix` to the token names were we don't have one in the current system.

As for @rezrah work, I already talked to him and this will be solvable through different means that are more typical for Style Dictionary.

> re: base scale- the only base scale we don't want to keep references for are base scale colors. Everything else is fair game.

Okay, that should be easy enough.

lukasoppermann:
Hey @rezrah @josepmartins @langermank, as you probably know, we have `color scales` per color mode and we have also exceptions that overwrite the `functional tokens` in parts.

Personally I find the most elegant way to be by using files for everything, as this is the default SD way, it needs no custom code and there is no "magic" you need to learn.

So my idea is:
1. We have some base functional files for color, like `primitives-light.json` or `component-light.json` (we can discuss how we split them in a separate space).
2. All color modes use those files. However If we have overwrites (not all colors do) we create a new file `component-light-high-contrast.json` which only holds the overwrites
3. The output is exactly how it is now and we have a clear separation of the exceptions

Any major concerns?

lukasoppermann:
Closing as everything is not merged via multiple other PRs.

PR #356, opened 2022-09-22
Add consumer test for primer brand

rezrah:
Adds a consumer / integration test for Primer Brand to help catch regressions end-to-end faster.

### Why do we need this?

We're currently refactoring the v2 token build scripts for product tokens. Primer Brand uses the same scripts to generate its own brand-specific tokens downstream, so the likelihood of regression there is high.

### How does it work?

This automation script will ephemerally build the canary package for Primitives and test it against Primer Brand in an Actions runner, to save time manually smoke testing each time a change is made.

Screenshot of GitHub Actions check in a pull request footer showing successful consumer test

### DIFF
--- a/.github/workflows/consumer_test.yml
+++ b/.github/workflows/consumer_test.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   test:
-    if: ${{ github.repository == 'primer/brand' }}
+    if: ${{ github.repository == 'primer/primitives' }}
     name: Primer Brand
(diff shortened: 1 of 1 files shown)


langermank (approved):
I'm not a workflow expert but I trust you! 😆 🚀

colebemis (approved):
Looks good 📦 🚢

lukasoppermann (approved):
Awesome.

PR #366, opened 2022-10-10
Style dictionary base and functional colors

lukasoppermann:
This PR adds:
- color design tokens in json format for both `base` and `functional` colors
- a new build script to convert the colors 

### Context
This is extracted from #353 to only add the colors and not change any of the existing tokens. This should make it easier to review and merge this PR.


config/formats/javascript-commonJs.ts
```diff
@@ -0,0 +1,24 @@
+
```
rezrah, line 1: stray newline

config/formats/javascript-export.ts
```diff
@@ -0,0 +1,24 @@
+
+import StyleDictionary from 'style-dictionary';
+import { format } from "prettier";
+import { jsonToNestedValue } from '../utilities/jsonToNestedValue';
+
+const { fileHeader } = StyleDictionary.formatHelpers;
+
+export const javascriptExport: StyleDictionary.Formatter = ({ dictionary, file, options, platform }) => {
```
rezrah, line 7: I can't tell the difference between this and the commonjs one 🙈. can we rename this to something more descriptive of what it is? Is this an ESM format?

lukasoppermann: Yes, it just uses `export` instead of `module.export` so I guess `esm` would be fine. 👍

config/formats/typescript-export-defition.ts
```diff
@@ -0,0 +1,145 @@
...
+ * @description extract types from designTokenTypes (string with type definitions from file) and replace quoted types "Color" with unquoted Color
+ * @param output 
+ * @param designTokenTypes 
+ * @returns string with unquoted types
+ */
+const unquoteTypes = (output: string, designTokenTypes: string[]): string => {
+  // regex to match types like `type Color`
+  // const findTypesRegex = `type ([A-Za-z]+)`
+  // // extract types from designToken
+  // const types = [...designTokenTypes.matchAll(new RegExp(findTypesRegex, "g"))].map(match => match[1])
```
rezrah, line 19: Are these commented out code or annotations. Can we remove them?

config/formats/typescript-export-defition.ts
```diff
@@ -0,0 +1,145 @@
...
+  return designTokenType
+}
+
+/**
+ * convertPropToType
+ * @description return take type attribute and return correct typescript type as string 
+ * @param type 
+ * @returns 
+ */
+const convertPropToType = (value: any, type: string): string => {
```
rezrah, line 45: Would value not always be string or number?

config/formats/typescript-export-defition.ts
lukasoppermann: Yes.

config/formats/typescript-export-defition.ts
lukasoppermann: No, it could be an object, e.g. for shadow. Although by this time it may already be converted to a string, I have to check this.

config/formats/typescript-export-defition.ts
```diff
@@ -0,0 +1,145 @@
...
+  if ('value' in token) {
+    usedTypes.add(convertPropToType(token.value, token.$type))
+    return convertPropToType(token.value, token.$type)
+  }
+  // is non-object value
+  if (typeof token !== 'object') return token
+  // is obj
+  const nextObj = {}
+  for (const [prop, value] of Object.entries(token)) {
+    // @ts-ignore
```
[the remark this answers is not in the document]
lukasoppermann: This was just too much for me 😅 

> config/formats/typescript-export-defition.ts:77:5 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
>   No index signature with a parameter of type 'string' was found on type '{}'.

config/formats/typescript-export-defition.ts
```diff
@@ -0,0 +1,144 @@
...
+  // clean up types
+  usedTypes.forEach((type) => {
+    if (!['ColorHex', 'Shadow', 'Size', 'SizeRem'].includes(type)) {
+      usedTypes.delete(type)
+    }
+  })
+  // get token type declaration from file
+  let designTokenTypes = []
+  for (const type of usedTypes) {
+    // @ts-ignore: Argument of type 'string' is not assignable to parameter of type 'never'.
```
rezrah, line 105: Does this go away if you cast designTokenTypes as a string array

config/tranformers/name-path-to-dot-notation.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @todo: replace with _.capitalize once we don't use camelCase in token names anymore
+ * @param param0 
+ * @returns 
+ */
+const capitalize = ([firstLetter, ...restOfWord]: string): string => {
+  return firstLetter.toUpperCase() + restOfWord.join("")
+}
+
+
+const camelCase = (string: string): string => {
```
rezrah, line 15: Should capitalize and camelCase be in this file? or hoisted out?

config/utilities/copyFilesAndFolders.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * copyFilesAndFolders
+ * @description Copies file from copyFilesAndFolders array this is used to copy deprecated and removed values
+ * @param filesAndFolders array of paths
+ * @param source path 
+ * @param destination path 
+ * @returns promise
+ */
+export const copyFilesAndFolders = async (filesAndFolders: string[], source: string, destination: string) => {
+  return Promise.all(filesAndFolders.map(
```
rezrah, line 11: Bit hard to read. Can you break this up into smaller units of code and compose them instead?

package.json
```diff
@@ -61,12 +62,15 @@
     "deep-shape-equals": "^0.1.2",
     "deepmerge": "^4.2.2",
     "flat": "^5.0.2",
+    "fs-extra": "^10.1.0",
```
rezrah, line 65: Do we need this? Node 16+ has much better file copying built-ins that we could use to avoid an extra dep here?

tempColorTokenBuild.ts
```diff
@@ -0,0 +1,135 @@
...
+import { jsonDeprecated } from './config/tranformers/json-deprecated';
+import { scssMixinCssVariables } from './config/formats/scss-mixin-css-variables';
+import { javascriptCommonJs } from './config/formats/javascript-commonJs';
+import { javascriptExport } from './config/formats/javascript-export';
+import { typescriptExportDefinition } from './config/formats/typescript-export-defition';
+import { platformJson } from './config/platforms/json';
+import { namePathToDotNotation } from './config/tranformers/name-path-to-dot-notation';
+
+const buildPath = 'tokens-v2-private'
+const prefix = 'primer'
```
rezrah, line 24: can we make this a system argument or env var? We'd need to customise this for Brand.

config/formats/typescript-export-defition.ts
lukasoppermann: Okay, I tested and at this point it is already a string, so I changed it, great catch. 👍

rezrah (commented):
This is nice work @lukasoppermann. On surface it LGTM but there's a lot going on.

Final feedback. / questions:

- Are you planning to add test coverage? Might be a good idea to do that now or straight after given how the code is more disparate. 
- Let's get linting and formatting in _before you merge this_
- What's the plan for making this interoperable with Primer Brand? The current steps seem to only cater for the product tokens, even locking in the prefix.

config/tranformers/name-path-to-dot-notation.ts
lukasoppermann: Well in this case they are special:

- capitalize does not, e.g. like the lodash version, convert everything to lowercase. It only changes the first letter of the word, which may be unexpected but is need for keeping out names like `onEmphasis` in tact.

- camelCase is very specific in that it only combines the parts separated by space, `-` or `_`

Because of this I think keeping it internal would be good. 

But once we have the naming convention figured out and `eslint` for json set up and can be sure we get only specific things we could look into replacing it with different functions.

Maybe we should rename `capitalize` to `ucfirst`?

tempColorTokenBuild.ts
lukasoppermann: Yes, sure. Do you have a best practise recommendation for that?

lukasoppermann:
> 1. Are you planning to add test coverage? Might be a good idea to do that now or straight after given how the code is more disparate.
> 2. Let's get linting and formatting in before you merge this
> 3. What's the plan for making this interoperable with Primer Brand? The current steps seem to only cater for the product tokens, even locking in the prefix.

@rezrah 

1. yes, I'd prefer to add this after merging.
2. Yes, please. Once eslint is merged.
3. Open for whatever you suggest. `env` sounds good.

package.json
lukasoppermann: If we can replace it easily in https://github.com/primer/primitives/pull/366/files/1113d574244fd445b1bbf6d216a81ee2c5256671#diff-19c3a8039a4d716c096554204131cf6aad8c75aec66e5756c2b0a0e9e5a2d939R12 I am all in.

I think we can now narrow it down to just copying folders with all files inside. This may make it easier, as it does not need to copy files as input. I can also look into to.

config/utilities/copyFilesAndFolders.ts
lukasoppermann: I extracted the `errorHandler` I hope this makes it easier to read.

As we said we once merged we should create an issue to replace this function including the `fse` with our own implementation of using `fs` to avoid the dependency.

package.json
lukasoppermann: See: https://github.com/primer/primitives/pull/366#discussion_r992307456

tempColorTokenBuild.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * convert colors
+ */
+for (const [theme, source, include] of themes) {
+  StyleDictionary.extend(getStyleDictionaryConfig(`color/${theme}`, source, include)).buildAllPlatforms()
+}
+// TODO: Remove once shadows that used to be in colors are implemented
+// eslint-disable-next-line no-console
+console.log('⚠️ Shadows are not implemented in tokens correctly')
```
langermank, line 136: Nice 👍

langermank (commented):
This is looking really great. I'm relying on @rezrah's technical expertise for final review 🙌 

Have we thought about a way we might be able to do a test/compare the values of the old color token set vs. this new one? Either compare the values, or visually render and compare (or maybe both).

tempColorTokenBuild.ts
lukasoppermann: > This is looking really great. I'm relying on @rezrah's technical expertise for final review 🙌
> 
> Have we thought about a way we might be able to do a test/compare the values of the old color token set vs. this new one? Either compare the values, or visually render and compare (or maybe both).

Hey, this is complicated without putting to much work into it.

What I have done is a simple `diff` of the old and new output files (`sass`, `css`, `js`).
However it requires some changes in the old code as the output there is a bit mixed.

I would suggest we do this in a separate PR that we never merge but just use to surface those diffs.
Also keep in mind that this only works until we start changing the names.

@rezrah do you have a better idea?

config/tranformers/name-path-to-dot-notation.ts
rezrah: Makes sense 👍 

In that case, can we add these as comments / annotations above the functions please? So it's clear to others these are non-standard?

Also agree on changing naming.. I'd recommend going verbose and not abbreviating anything. E.g. `lowerCaseFirstCharacter`.

tempColorTokenBuild.ts
lukasoppermann: Not a big fan of this, but let's to it for now to get moving and see if we can improve this later.

config/tranformers/name-path-to-dot-notation.ts
lukasoppermann: Done.

package.json
```diff
@@ -28,13 +28,14 @@
...
-    "format": "prettier --check '**/*.{js,jsx,ts,tsx,md,mdx,css,json}'",
+    "format": "prettier --check '**/*.{js,jsx,ts,tsx,md,mdx,css}'",
     "lint": "eslint '**/*.{js,ts,tsx,md,mdx}' --max-warnings=0 --config .eslintrc.js",
     "postbuild": "ts-node ./script/color-contrast.ts",
     "prebuild": "rm -rf dist",
     "prebuild:tokens": "rm -rf tokens-v2-private",
-    "prepack": "yarn build && yarn build:tokens",
+    "prepack": "yarn build && yarn build:tokens && yarn tempColorTokens",
     "release": "changeset publish",
+    "tempColorTokens": "ts-node tempColorTokenBuild.ts",
```
rezrah, line 38: sorry, missed this one before.. one final request for consistency..

```suggestion
    "build:color-tokens": "ts-node tempColorTokenBuild.ts",
```

rezrah: also, can we move this file to `scripts/` folder? Only reason `build.js` isn't, is because it's surfaced in the npm package.

tempColorTokenBuild.ts
```diff
@@ -0,0 +1,172 @@
...
+  //
+  for (const [name, source, include] of deprecatedBuilds) {
+    //
+    StyleDictionary.extend(deprecatedConfig(name, source, include, buildOptions)).buildAllPlatforms()
+  }
+}
+
+buildDesignTokens({
+  buildPath: 'tokens-v2-private',
+  prefix: undefined
```
rezrah, line 171: Should this be 'primer'?

rezrah (approved):
LGTM. Some additional minor feedback left, but I'll 👍 for now.

tempColorTokenBuild.ts
rezrah: Also, on the build path, I'm guessing it's intentional but the colors output in a different folder to the other existing v2 tokens. 
Screenshot 2022-10-12 at 15 36 58

lukasoppermann: Yes, this is intentional. I see no reason why we should have a another level of nesting and @langermank didn't reply / disagree when I mentioned to remove this somewhere a few weeks ago. Ultimately we should move all tokens out of the tokens subfolder.

> Should this be 'primer'?
not for not, to keep parity, we have no prefix anywhere.

package.json
lukasoppermann: We can do this for now. In the end this will be integrated / updating `build.ts`.

tempColorTokenBuild.ts
lukasoppermann: Currently our `color` tokens are not, this is why the prefix is set to `undefined`.

If we set it to `primer` it would be prefixed. However this is something we can define when we compared the outputs.
It is just on option.

tempColorTokenBuild.ts
rezrah: > Not for now, to keep parity, we have no prefix anywhere.

Oh i see. I thought this was one of the main distinctions with v2 tokens, that we always namespace them. It was also to help avoid confusion when brand tokens are available alongside the the product tokens.

tempColorTokenBuild.ts
lukasoppermann: I think I just don't like that it is a giant function  and would prefer to provide configs via something like env or a file. 

But this is I guess personal taste. 😅

lukasoppermann:
I removed this from the `prepack` script so color tokens will not be available in the package. This should allow us to merge this but continue working on fine-tuning stuff like naming, etc.

I need this to be merged as a platform to add the shadows, etc. but it does not need to be visible to users, not even in the private folder. So this should now be safe to merge with no side-effects.

tempColorTokenBuild.ts
lukasoppermann: I am not sure, I am all in for doing this. If we want to, we can just add it here. 

I am fine either way, it will just mean **all tokens** don't match and **all** will have to be replaced. If this is a definite, adding it in is literally changing this one line of code.

I think this should not be part of this PR as it is a longer discussion including if `ts` and `js` also have this because you import from something and name it again.

I believe for this PR (as it does not produce any output now) it is enough that the possibility is available to make it work either way.

Do you agree?

langermank: > tokens sub folder

I'm fine with removing this `tokens` sub folder, but I think base scales should be inside `/base` and functional colors should be inside `/functional`. I think you're building all themes (base scale and functional) into a single file, which I'm not sure about.. it feels like they should be separated the same way the base/functional sizing tokens are.

> primer prefix

I'm okay with leaving it off for now since we're not shipping these tokens yet, but it will be required for all the renames.

lukasoppermann: > I don't consider this merge blocking so long as we can keep iterating from here 👍

Yes, we can iterate on this until it is perfect.

> I'm okay with leaving it off for now since we're not shipping these tokens yet, but it will be required for all the renames.

I understood this a lot better now in our zoom and will work on getting the naming right in a new PR were we can only focus on `folder structure` and `naming/prefix`.

langermank (approved):
I think this is good to merge so we can keep iterating!

PR #367, opened 2022-10-10
Add code linting and formatting

rezrah:
Adding ESLint and Prettier for improved DX and quality control.

### What's changed
- Added required configurations for linting and formatting
- Fixed all errors that subsequently appeared
- Added a CI check to catch future errors relating to QA

#### Locally

- clone branch
- `yarn install --frozen-lockfile`
- `yarn lint`
- `yarn format`

#### In CI

- Scroll down to Checks in this PR
- Find Action results for `CI / Install and run tests`
- Verify it's 🟢 
  CI check showing as passed

- Verify there's no design tokens being changed

### Out of scope in this PR
- Adding linting and formating fixes for v1 tokens. Too many errors to fix for code that is due to be sunsetted soon.


script/color-contrast.ts
```diff
@@ -97,76 +104,80 @@ const testContrast = (minimumContrast: number, colorA: string, colorB: string, b
...
   console.log(contrastTable.render())
 }
+
 /**
  * run tests, output results to console and store them for json
  */
-const results = Object.entries(contrastRequirements)
-  .map(([theme, colorPairs]: [theme: string, colorPairs: ContrastRequirement[]]) => {
-    // turn deeply nested colors object into one level object like 'fg.default': '#000'
-    // @ts-ignore
```
rezrah, line 154: @lukasoppermann - I removed the ts-ignore here by switching to an unnamed tuple. You may wanna check you're happy with this.

script/color-contrast.ts
```diff
@@ -97,76 +104,80 @@ const testContrast = (minimumContrast: number, colorA: string, colorB: string, b
...
-  })
+const results = Object.entries(contrastRequirements).map(([theme, colorPairs]: [string, ContrastRequirement[]]) => {
+  // turn deeply nested colors object into one level object like 'fg.default': '#000'
 
-// write json file for workflow
-fs.writeFile('dist/color-contrast-check.json', JSON.stringify(results), (err) => {
-  if (err) throw err;
-  console.log('The file has been saved!');
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const flattenColors = flattenObject((colors.colors as any)[theme])
```
rezrah, line 165: To patch the typescript error, I've cast this any for now. Hopefully we won't have this with new tokens that have their own type definitions. This is an example of where we'd want stricter typing of our JSON @lukasoppermann.

tsconfig.build.json
```diff
@@ -0,0 +1,24 @@
+{
```
rezrah, line 1: This file is for the v1 tokens. I've renamed this to tsconfig.build.json as it was causing compilation issues for everything else that uses TS.

langermank (approved):
Thank you so much Reza! Looks good to me 😄

tsconfig.build.json
lukasoppermann: 👍

script/color-contrast.ts
```diff
@@ -30,56 +30,62 @@ const getOpaqueColor = (color: string, background: string): string => {
...
-        ...testContrast(minimumContrast, colors[colorA], colors[colorB], undefined, contrastPair),
+        ...testContrast(minimumContrast, scopedColors[colorA], scopedColors[colorB], undefined, contrastPair),
         minimumContrastRatio
       }
     }
-    // if colorB is semi-transparent 
-    // get the correct canvas colors to test agains
-    let canvasColorArrays = cavnasColors
+    // if colorB is semi-transparent
+    // get the correct canvas colors to test again
```
lukasoppermann, line 60:
```suggestion
    // get the correct canvas colors to test against
```

script/color-contrast.ts
```diff
@@ -30,56 +30,62 @@ const getOpaqueColor = (color: string, background: string): string => {
...
  * @description runs through all color pairs of a theme and checks the contrasts
- * @param colorPairs 
- * @param colors 
+ * @param colorPairs
+ * @param colors
  * @returns contrastTestResult
  */
-const runContrastTest = (colorPairs: ContrastRequirement[], colors: any): contrastTestResult[] =>
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const runContrastTest = (colorPairs: ContrastRequirement[], scopedColors: any): contrastTestResult[] =>
```
lukasoppermann, line 44: Why did you rename it to `scopedColors` colors? I agree that `colors` is not the best and maybe `scopedColors` is great, I just don't really understand the name. Can you please explain it to me? 😬

script/color-contrast.ts
lukasoppermann: Yes, but we would have to use `string` or `colorHex` with the new tokens, right? As we can't add types to the json directly. So we still need to cast, just not `any`.

script/color-contrast.ts
rezrah: thanks for reminding me, I forgot to annotate this one. 

colors is already mentioned in the upper scope, where you're importing the tokens from dist. linters picked this up, so I've renamed this scopedColors to imply it's locally scoped to the function closure.

script/color-contrast.ts
rezrah: ah right.. was trying to fix a typo and ended up creating another one 😅

script/color-contrast.ts
rezrah: yes, any is here to help remove the `@ts-ignore` that was previously here. Ideally this JSON source would be strongly typed like an API contract, which we can cast immediately after importing.

lukasoppermann: Is this possible to make the JSON strongly type or do you mean by enforcing it through `eslint`?

rezrah: I imagine it to work like an API response, where we can provide a corresponding typed interface that is co-located to the JSON output, which either tsc can pick up at the same time or we import manually as cast against the import.

lukasoppermann: Did you do this before? I am not really sure how this would work. Would we just hand-type a file and assume in `colors/*.json` are only color tokens. Or we could allow all types of tokens.

I am happy to work on this but it would be great if you could provide an example here so I am moving in the right direction.

rezrah: Yeah it's essentially the same thing style-dictionary already does for its own built-in typed interfaces. Instead of using it for CommonJS modules we use it for JSON. The Style Dictionary version is actually just stringified JSON, so it'll work out of the box. I can show you an example of this later separately.

PR #368, opened 2022-10-14
Add transformers and formatters to SD end export a preset

lukasoppermann:
In an effort to clean up the build file I created a preset which does the following:

- import style dictionary
- register parser
- register formats
- register transformers

Exports style dictionary with the registered transformers

This removes all the setup work from the build script. Since the `style dictionary` object is exported it can still be configured more so brand can add custom transformers and we can add platforms etc.

Missing / problems:
- [ ] I would like to pipe the `StyleDictionary` type definitions through, so there is no need to import style dictionary in build
- [ ] TSDoc links to files are relative to the preset and don't work in other files (can this be fixed)?

### DIFF
--- a/script/tempColorTokenBuild.ts
+++ b/script/tempColorTokenBuild.ts
@@ -104,8 +104,6 @@ export const buildDesignTokens = (buildOptions: ConfigGeneratorOptions): void =>
    * convert colors
    */
   for (const [theme, source, include] of themes) {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
     PrimerStyleDictionary.extend(
       getStyleDictionaryConfig(`color/${theme}`, source, include, buildOptions)
     ).buildAllPlatforms()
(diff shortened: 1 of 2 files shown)


simurai (commented):
Does this change anything with the current output?

lukasoppermann:
> Does this change anything with the current output?
No. It does not change any output, also not the new. It is just a restructuring of the build process.

simurai (approved):
> No. It does not change any output

Ok, I can't quite judge the changes, but as long as the output stays the same, this seems low risk.

script/tempColorTokenBuild.ts
```diff
@@ -127,7 +104,9 @@ export const buildDesignTokens = (buildOptions: ConfigGeneratorOptions): void =>
    * convert colors
    */
   for (const [theme, source, include] of themes) {
-    StyleDictionary.extend(
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
rezrah, line 108: What was the error here?

lukasoppermann: Was a leftover from before I think.

PR #369, opened 2022-10-14
Fix consumer test for primer brand

rezrah:
Primer Brand has been converted to a monorepo. The commands required to trigger a token build in that repo have been altered. 

This PR fixes the test by adjusting to the new monorepo configuration.


lukasoppermann (approved):
LGTM

PR #371, opened 2022-10-14
Exporting tokens as custom-properties json object

jonrohan:


lukasoppermann:
Hey @jonrohan,

this is interesting for me. I talked with @langermank about it today. Is this part of exploring a path for PRC to use `css variables`?

As I am building the new Style dictionary build I would love to know:
- what is the output you would like to see?
- how would a token reference another token in this output format? Using `css variables` as well?
- is this just an experiment or is there a consensus as to what the PRC team needs?

jonrohan:
> Is this part of exploring a path for PRC to use `css variables`?

This is in support of inlining fallbacks in the CSS we write in PVC https://github.com/primer/view_components/pull/1453

> * what is the output you would like to see?

Basically I need json files like this, with `cssvar: value`

```json
{
  "customProperties": {
    "--base-size-4": "4px",
    "--base-size-8": "8px",
    "--base-size-12": "12px"
  }
}
```

> * how would a token reference another token in this output format?

I'm running into this problem right now, In this case I think I need the original value because the intent of the plugin in PVC is to protect when the variables don't exist.

```
{
  "customProperties": {
    "--primer-borderInset-thin": "inset 0 0 0 {primer.borderWidth.thin}",
```

> * is this just an experiment or is there a consensus as to what the PRC team needs?

Not for PRC, for PVC

lukasoppermann:
> I'm running into this problem right now, In this case I think I need the original value because the intent of the plugin in PVC is to protect when the variables don't exist.
> 
> {
>   "customProperties": {
>     "--primer-borderInset-thin": "inset 0 0 0 {primer.borderWidth.thin}",

So from the PVC perspective it would be better to keep `{primer.borderWidth.thin}` in there?
Or to use a css variable:

```css
 "--primer-borderInset-thin": "inset 0 0 0 var(--primer-borderWidth-thin)",
```

Or did you mean to avoid references all together:

```css
 "--primer-borderInset-thin": "inset 0 0 0 1px",
```

jonrohan:
Closing this for now. Still need a more complete export of variables that exist but not sure my pr is helpful there.

PR #372, opened 2022-10-17
Adding tests for build process

lukasoppermann:
## Summary
This PR adds tests for all or most functions within the `config` folder. All those functions are used when building color design tokens and are essential for the build so we should make sure they work as expected.

## Notable changes
- [x] adds tests for parser
- [x] adds tests for transformer
- [x] adds tests for filter
- [x] adds tests for utilities
- [x] adds tests for formats


__tests__/config/formats/typescript-export-definition.test.ts
```diff
@@ -0,0 +1,124 @@
+import {typescriptExportDefinition} from '../../../config/formats/typescript-export-defition'
```
rezrah, line 1: Heads up, typo in the original filename. 

```suggestion
import {typescriptExportDefinition} from '../../../config/formats/typescript-export-definition'
```

__tests__/config/formats/typescript-export-definition.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        yellow: {
+          value: '#FF0000'
+        }
+      },
+      size: {
+        default: {
+          value: 16,
+          $type: 'dimension'
+        },
+        // px: {
```
rezrah, line 20: Are you planning to remove these?

__tests__/buildJs/pathToDotNotation.test.js
```diff
@@ -1,4 +1,4 @@
-const pathToDotNotation = require('../build').pathToDotNotation
+const pathToDotNotation = require('../../build').pathToDotNotation
```
rezrah, line 1: general feedback / question here, but could we start co-locating these tests to the source files? We do this in our other Primer repo's, and think it should apply here too.

__tests__/config/formats/typescript-export-definition.test.ts
lukasoppermann: No I am planning on re-enabling them once we move the size stuff over to this build process.

__tests__/buildJs/pathToDotNotation.test.js
lukasoppermann: Those three tests will be removed once we move everything from `build.js` into the new build file.

So if we talk about colocation it would be for the config folder. I am not 100% sure I love the idea, as this either means we have double the files e.g. `16` instead of `8` which makes folders a bit hard to parse or, we need to put every transformer, etc. into a subfolder just to co-locate the test, which also makes it hard to work.

With components it feels more natural as you have the css module there as well.

Maybe we can discuss this in a call as this may be a bit of a longer conversation. 😉

rezrah: yeah, I can see why you'd want to keep these build.js specific tests here for now 👍 . I think the others _could_ move though. E.g. `__tests__/config/formats/typescript-export-definition.test.ts` can be colocated to `config/formats/typescript-export-definition.ts` and there's _less_ folder creation and traversal and those tests still pass fine. Might be missing something here, so happy to sync on it.

__tests__/buildJs/pathToKebabCase.test.js
```diff
@@ -1,4 +1,4 @@
-const pathToKebabCase = require('../build').pathToKebabCase
+const pathToKebabCase = require('../../build').pathToKebabCase
```
rezrah, line 1: While we're doing this.. can we switch to restructuring rather than assignment through notation?

```suggestion
const {pathToKebabCase} = require('../../build')
```

__tests__/config/filters/isSource.test.ts
```diff
@@ -0,0 +1,21 @@
...
+      value: 'blue is not source',
+      isSource: false
+    },
+    {
+      value: 'yellow is not source'
+    }
+  ] as StyleDictionary.TransformedToken[]
+  it('filters out non-source items', () => {
+    expect(items.filter(isSource)).toStrictEqual([items[0]])
+  })
```
rezrah, line 20: A suggestion to improve tests like this.. 

You're asserting against a very specific position in the test fixture, but that value can change at any time and invalidate the test. It also doesn't capture unhappy paths or edge cases where multiple  truthy results are present in the fixture. 

```suggestion
  const result = items.filter(isSource)
  const expectedValue = items.filter(item => item.isSource)
  
  it('filters out non-source items', () => {
    expect(result).toStrictEqual(expectedValue)
  })
```

This should still achieve the same thing but now isn't dependent on the fixture data being correct, but also keeps `isSource` honest against regressions.

Ideally this would be done with property testing paradigms, but this should work in the interim.

__tests__/config/formats/javascript-commonJs.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it('Formats tokens without prefix', () => {
+    const output = `exports.default = {
+  colors: {
+    red: "#FF0000",
+  },
+};
+`
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: fake values to test formatter
```
rezrah, line 48: Worth stubbing these out? Could make them available globally as we seem to be using these in a few different tests.

__tests__/config/parsers/w3c-json-parser.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      color: {
+        comment: 'a red color',
+        value: 'red',
+        $type: 'color',
+        alpha: 0.5
+      }
+    })
+  })
+
+  it('parses empty json', () => {
```
rezrah, line 28:
```suggestion
  it('throws an error while parsing JSON without required properties', () => {
```

__tests__/config/parsers/w3c-json-parser.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      })
+    ).toStrictEqual({
+      color: {
+        comment: 'a red color',
+        value: 'red',
+        $type: 'color',
+        alpha: 0.5
+      }
+    })
+  })
```
rezrah, line 26: nit for readability

```suggestion
 it('parses valid json5', () => {
    const mockJSON = {
      contents: `{
        "color": {
          // comment
          "$value": 'red',
          "$description": "a red color",
          "$type": "color",
          alpha: 0.5
        }
      }`,
      filePath: ''
    }
    
    const result = w3cJsonParser.parse(mockJSON)

    const expectedOutput = {
      color: {
        comment: 'a red color',
        value: 'red',
        $type: 'color',
        alpha: 0.5
      }
    }

    expect(result).toStrictEqual(expectedOutput)
  })
```

__tests__/config/formats/javascript-commonJs.test.ts
lukasoppermann: Yes, we could. Either we create a dataset somewhere or we create a function that adds all the other stuff to the tokens so you only define what is needed for the test. I'll look into it.

config/formats/scss-mixin-css-variables.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      & {
+        --red: transformedValue;
+      }
+    }`,
+      {parser: 'scss', printWidth: 500}
+    )
+    expect(scssMixinCssVariables(input)).toStrictEqual(expectedOutput)
+  })
+
+  it('Formats throws error if options.mixinName is missing', () => {
```
rezrah, line 25: should this be:

```suggestion
  it('throws and error if options.mixinName is missing
```

config/formats/typescript-export-definition.test.ts
```diff
@@ -0,0 +1,90 @@
+import {typescriptExportDefinition} from './typescript-export-definition'
+import {getMockFormatterArguments} from '../../test-utilities/getMockFormatterArguments'
+import {format} from 'prettier'
+import {getMockDictionary} from '../../test-utilities/getMockDictionary'
+import {getMockToken} from '../../test-utilities/getMockToken'
+
+describe('Format: Tyepscript definitions', () => {
```
rezrah, line 7:
```suggestion
describe('Format: TypeScript definitions', () => {
```

config/formats/typescript-export-definition.ts
```diff
@@ -36,22 +43,24 @@ const getTokenType = (tokenTypesPath: string): string => {
...
     case 'color':
       if (value[0] === '#') {
         return 'ColorHex'
       }
       return 'string'
     case 'dimension':
-      if (value.substring(value.length - 3) === 'rem') {
-        return 'SizeRem'
-      }
+      if (typeof value === 'number') return 'number'
```
rezrah, line 54: value seems to be typed as a string.. should this be !isNan instead? to account for values like "2"? or should the type defs be updates to accept string and number.

config/utilities/jsonToNestedValue.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        noValue: {
+          string: 'string',
+          number: 12
+        }
+      }
+    }
+  }
+
+  it('it returns only value property as value of object', () => {
+    expect(jsonToNestedValue(obj['parent']['subparent']['child1'])).toStrictEqual('#111')
```
rezrah, line 28: just curious.. do you need to use index here? makes it a little harder to read... this should work same...

```suggestion
   expect(jsonToNestedValue(obj.parent.subparent.child1)).toBe('#111')
```

config/utilities/upperCaseFirstCharacter.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('Utilities: upperCaseFirstCharacter', () => {
+  it('it transforms all lowercase word', () => {
+    expect(upperCaseFirstCharacter('primer')).toStrictEqual('Primer')
+  })
+
+  it('it transforms all lowercase sentence (words with spaces)', () => {
+    expect(upperCaseFirstCharacter('primer design token')).toStrictEqual('Primer design token')
+  })
+
+  it('it transforms all uppercase word', () => {
```
rezrah, line 12: is this a more accurate test name?
```suggestion
it('it preserves casing for words that are already all uppercased', () => {
```

tsconfig.json
```diff
@@ -13,6 +13,6 @@
     "declaration": true,
     "resolveJsonModule": true
   },
-  "include": ["src/", "script/", "docs/", "config/types/"],
+  "include": ["src/", "script/", "docs/", "config/types/", "config/**/*.test.ts", "test-utilities/"],
```
rezrah, line 16: could test-utilities go under src instead? This `includes` array is getting quite large and might make it less predictable for others when it comes to debugging.

config/formats/typescript-export-definition.ts
lukasoppermann: I think updating the type definition would be correct here. But technically a `[dimension` must be a string with `px` or `rem`](https://design-tokens.github.io/community-group/format/#dimension) so maybe we can just scrap this line. What do you think?

rezrah: yeah scrap it if it never gets evaluated 👍

rezrah (approved):
Lovely work as always @lukasoppermann ✨ 

Left some minor feedback for you, but otherwise good to go 🚀

tsconfig.json
lukasoppermann: Yes, does that make sense? I think we wanted to [move tokens into source (`src/`)](https://github.com/github/primer/pull/1309/files?short_path=722b130#diff-722b130da8c969a084abdcb34977e310c071d1af7f0458ef6f8a35a7155c1a4c) so would this still make sense then? And if yes would we move `script/` into source as well? And `config`?

Is there a common way to structure this at github?

rezrah: not an enforced convention AFAIK, but you typically not want to overload the includes array and rather use it as a catch-all. Also you have config/types in there. Not sure what that is exactly, but just on naming it may be worth putting that under the [typeRoots](https://www.typescriptlang.org/tsconfig#typeRoots) property instead?

PR #377, opened 2022-10-19
Better naming and structure for token files

lukasoppermann:
## Summary

- Refactors the arrays that store the token files to be processed by SD from a tuple to an object of the type [`TokenBuildInput`](https://github.com/primer/primitives/compare/improve-build?expand=1#diff-360f85412ee892a67037aca00113357cade471fbbd9d84c7e774b8ec3aa729f2)
- moved base color token files into `dark` and `light` folders

## List of notable changes:

- **added** [`TokenBuildInput`](https://github.com/primer/primitives/compare/improve-build?expand=1#diff-360f85412ee892a67037aca00113357cade471fbbd9d84c7e774b8ec3aa729f2) and [`TokenBuildInputArray`](https://github.com/primer/primitives/compare/improve-build?expand=1#diff-360f85412ee892a67037aca00113357cade471fbbd9d84c7e774b8ec3aa729f2)
- **moved** all base dark json files into `base/dark` and all light files into `base/light`
- **renamed** all modes that are not default to be prefixed with `extension.` to show that they extend the base mode (`dark` or `light`)
- **changed** base files to be `.json5` to be able to add a comment that states that the file extends another file

- do you understand the new folder structure in `base/color`
- should we do the same for `functional/color`?
- is there a better name for the `TokenBuildInput` type? Or a better way to store those inputs?

## Steps to test:

1. clone repo
1. `npm i --legacy-peer-deps`
1. `npm run build:color-tokens`

## Contributor checklist:

- [x] All new and existing CI checks pass
- [x] Related issues have been referenced in the PR description


lukasoppermann:
@rezrah I think the consumer test fails because I am using `.json5`. Can I send a PR to brand to make `json5` work with brand so it does not break?

simurai (commented):
> do you understand the new folder structure in base/color
> should we do the same for functional/color?

Currently a token like `{base.color.scale.gray.9}` can be found in `base/color/`, which is easy to find. But then for functional colors like `{color.fg.default}` it's hard to know where that token comes from unless you know it's inside `functional/color/`.

I think we already talked about it, but I wonder if we really should make `base` and `functional` the top level folders? Currently [tokens/base/typography/](https://github.com/primer/primitives/blob/9d481935acb95dd3a177854410a13a24593a6939/tokens/base/typography/typography.json) only has 4 weights, but that could also just be in `tokens/typography/weight.json` together with the rest of `typography`. In other words, maybe not each group (color, size, typography) needs the two folders called `base` + `functional` and can be organized however it makes most sense.

```
tokens
    > color
        > base
        > functional
    > size
        - base.json
        - border.json
        - breakpoints.json
        - etc.
    > typography
        - code.json
        - font-stack.json
        - text.json
        - weight.json
```

And to add to that, maybe a simple rule could be: The **name of a token** should always **match the folder structure**. So that it easily can be found. E.g.

- Instead of `control-minTarget-fine` being in `functional/size/size.json`
- it should be called `size-control-minTarget-fine` and be found in `size/control.json`

I know that would mean a lot more changes. 🙈

lukasoppermann:
Hey @simurai,

the change in this PR is actually just moving the base color tokens (and this should also be done for functional tokens) int a `dark` and a `light` folder and prefixing the themes with `extension.`. So I think your comment should not hold up this PR.

However I find this to be a very valid and interesting point, maybe we should move this discussion into [the current ADR](https://github.com/github/primer/pull/1309/files)? Do you want to post it over there?

I haven't thought through your suggestion for very long yet, but I do like inverting the folders as this brings the tokens that belong together closer together.

I am not 100% sure if we really want to remove the `base` and `functional` folders in some cases, I currently prefer to keep the same structure for all folders to avoid confusion.

simurai:
> So I think your comment should not hold up this PR.
> maybe we should move this discussion into [the current ADR](https://github.com/github/primer/pull/1309/files)? Do you want to post it over there?

Sounds good. Yeah, this was more a general drive-by comment.

@types/TokenBuildInputArray.d.ts
```diff
@@ -0,0 +1,10 @@
+type TokenBuildInputItem = {
```
rezrah, line 1: is @types folder new? If so, can we move this to a non-namespaced folder to avoid confusion with @types npm package namespace?

script/tempColorTokenBuild.ts
```diff
@@ -10,71 +10,79 @@ import {platformJs} from '../config/platforms/javascript'
...
+      include: [`tokens/base/color/dark/dark.json5`, `tokens/base/color/dark/extension.protanopia-deuteranopia.json5`]
+    },
+    {
+      outputName: 'dark-high-contrast',
+      source: [
+        `tokens/functional/color/dark/*.json5`,
+        `tokens/functional/color/dark/exceptions/exceptions.high-contrast.json5`
+      ],
+      include: [`tokens/base/color/dark/dark.json5`, `tokens/base/color/dark/extension.high-contrast.json5`]
+    }
```
rezrah, line 85: This is a lot clearer to me now.. thanks

@types/TokenBuildInputArray.d.ts
```diff
@@ -0,0 +1,10 @@
...
+  // Used as the filename for the generated file (e.g. `dark`.css), extension will be appended
+  outputName: string
```
rezrah, line 3: could this be filename instead? or is it used for something else too? Output can mean different things in style dictionary and build process, so filename feels more explicit.

@types/TokenBuildInputArray.d.ts
lukasoppermann: You mean using `types` ? I thought this would be beneficial in @types as a consumer would maybe like to use the type when they define their own inputs?

But of course at the moment there is no way to get to them as they are internal to the function.

@types/TokenBuildInputArray.d.ts
```diff
@@ -0,0 +1,10 @@
...
+  // Array of `filepaths` to token files that should be converted and included in the output
```
rezrah, line 4:
```suggestion
  // Array of `filepaths` to token files that should be converted and included in the output. Accepts relative or glob paths.
```

@types/TokenBuildInputArray.d.ts
lukasoppermann: Yes, it is used for other stuff sometimes. I was afraid that people would add the extension if we call it filename. So I need just the name, no extension. Would you still name it `filename`?

lukasoppermann: Maybe we could also go for `basename`? https://stackoverflow.com/questions/280767/definition-of-filename

rezrah: I see you changed to outnameName, that's definitely an improvement. Happy to stick with that. I still think filename is clearest IMO, also because in JS, the parse() fn will return `name` as the filename sans exension. Basename also works, but also probably requires some assumed knowledge there?

@types/TokenBuildInputArray.d.ts
rezrah: Sounds good. I was more referring to calling the folder `@types`, which is a little confusing to me because that's usually for local overrides or stubbing of node_modules. E.g from PRC: https://github.com/primer/react/tree/main/%40types

I was suggesting moving this to something like `/src/types/build-helpers`.. and then reexport from an index.d.ts if you want consumers to use it through the npm package.

PR #379, opened 2022-10-27
New token structure idea [Prototype, no not merge]

lukasoppermann:
## Summary
After talking to @rezrah and getting @simurai [comment](https://github.com/github/primer/pull/1309#discussion_r1004434657) on the structure I wanted to try out separating the components into a separate folder.

I also tried out the token naming scheme suggested in @simurai [comment](https://github.com/github/primer/pull/1309#discussion_r1004434657).

### Benefits are:
**folder structure:**
- strong separation of component and functional tokens
- we can have one file per component which makes it easier to see which component overwrites exists and could help in an effort to remove more and more component tokens (e.g. many files may have only one overwrite)

**naming convention:**
- the name `ref` (reference) tokens makes it more clear than `base` that those tokens are not really meant to be used directly.
- I think `sys` (system) tokens is a good name for our primitives, although we could also name them `prim` or `func` or `fun` 🤣.
- `comp` tokens are great to separate the overrides that are supposed to only be used by a specific component.
- we could also add another type like `gh` or `gui` or `prod` which would be similar to `brand`. The `gh` tokens would be stuff like `ansi` colors or `codemirror` styles that may not be specific to a component but are also not `sys` or functional colors (e.g. `--primer-gh-color-ansi-black`)

## List of notable changes:

- **added** new folder `tokens/component/` with some individual files (I have not separated all tokens yet)
- **updated** how prefixes work, now **ALL** tokens get the `--primer` prefix (we could also omit this completely for brevity or shorten to `pr`)
- **updated** sub-prefixes:
  - `base` tokens are now `ref` tokens
  - functional system tokens are `sys` tokens (= tokens that make up the design system)
  - functional component tokens are `comp` tokens
  - brand tokens could have a `brand` sub-prefix (`--primer-brand-color-fg-default`)

### Output
After this change in the `base` files we have something like:

```css
  --primer-ref-color-scale-black: #010409;
  --primer-ref-color-scale-white: #ffffff;
  --primer-ref-color-scale-transparent: #00000000;
  --primer-ref-color-scale-gray-0: #f0f6fc;
  --primer-ref-color-scale-gray-1: #c9d1d9;
  --primer-ref-color-scale-gray-2: #b1bac4;
  --primer-ref-color-scale-gray-3: #8b949e;
```

All other tokens are combined within the functional folder per color theme:

```css
  --primer-comp-colors-avatar-bg: #ffffff1a;
  --primer-comp-colors-avatar-border: #f0f6fc1a;
  --primer-comp-colors-counter-border: #00000000;
  --primer-comp-colors-topic-tag-border: #00000000;
  --primer-comp-color-selectMenu-backdrop-border: #484f58;
  
  --primer-sys-color-fg-default: #c9d1d9;
  --primer-sys-color-fg-muted: #8b949e;
  --primer-sys-color-fg-subtle: #6e7681;
  --primer-sys-color-fg-on-emphasis: #ffffff;
```

- does the folder structure make sense and make it easier?
- does the addition of `comp` tokens make sense?
- is the new sub-prefix structure better than the old? Does changing this have any side-effects?

## Steps to test:

1. pull down branch
1. look into `tokens/` folder
1. run `npm run build:color-tokens` to build css


simurai:
> does the folder structure make sense and make it easier?

If we use `ref/sys` in the token name, like `--primer-ref-color-scale-black: #010409;` shouldn't the folder name be called the same?

- rename `tokens/base/` to `tokens/reference/`
- rename `tokens/functional/` to `tokens/system/`

otherwise people might wonder "why is it called `ref` if the tokens can be found under `base`?".

> does the addition of comp tokens make sense?

Yes, we kinda need it for these [component tokens](https://github.com/primer/primitives/blob/main/data/colors/vars/component_light.ts). Ideally we can reduce them as much as possible, but we probably always need some tokens specific to a single component.

> is the new sub-prefix structure better than the old? Does changing this have any side-effects?

I like `system` over `functional`. And yes, `reference` maybe makes it even more clear that these tokens shouldn't be used directly.

> we could also add another type like `gh` or `gui` or `prod` which would be similar to brand. The `gh` tokens would be stuff like `ansi` colors or `codemirror` styles that may not be specific to a component but are also not `sys` or functional colors (e.g. `--primer-gh-color-ansi-black`)

An alternative to `gh` as "the rest" or "it doesn't fit anywhere else" might be `eco`? As in "ecosystem". These tokens are used more rarely, but are still part of the overall ecosystem. e.g. `--primer-eco-color-ansi-black`

> * **updated** how prefixes work, now **ALL** tokens get the `--primer` prefix (we could also omit this completely for brevity or shorten to `pr`)

Yes, my vote goes to the shorter `pr` (or `gh`). Because the prefix will be used EVERYWHERE, I don't see the benefit of spelling it out. And token names are already quite long so having a shorter prefix helps a bit in reducing that.

simurai (commented):
> does the folder structure make sense and make it easier?

If we use `ref/sys` in the token name, like `--primer-ref-color-scale-black: #010409;` shouldn't the folder name be called the same? So it would be:

- rename `tokens/base/` to `tokens/reference/`
- rename `tokens/functional/` to `tokens/system/`

otherwise people might wonder "why is it called `ref` if the token can be found under `base`?".

> does the addition of comp tokens make sense?

Yes, we kinda need it for these [component tokens](https://github.com/primer/primitives/blob/main/data/colors/vars/component_light.ts). Ideally we can reduce them as much as possible, but we probably always need some tokens specific to a single component.

> is the new sub-prefix structure better than the old? Does changing this have any side-effects?

I like `system` over `functional`. And yes, `reference` maybe makes it even more clear that these tokens shouldn't be used directly and only referenced in system tokens.

> we could also add another type like `gh` or `gui` or `prod` which would be similar to brand. The `gh` tokens would be stuff like `ansi` colors or `codemirror` styles that may not be specific to a component but are also not `sys` or functional colors (e.g. `--primer-gh-color-ansi-black`)

An alternative to `gh` as "the rest" or "it doesn't fit anywhere else" might be `eco`? As in "ecosystem". These tokens are used more rarely, but are still part of the overall ecosystem. e.g. `--primer-eco-color-ansi-black`

> * **updated** how prefixes work, now **ALL** tokens get the `--primer` prefix (we could also omit this completely for brevity or shorten to `pr`)

Yes, my vote goes to the shorter `pr` (or `gh`). Because the prefix will be used EVERYWHERE, I don't see the benefit of spelling it out. And token names are already quite long so having a shorter prefix helps a bit in reducing that.

lukasoppermann:
> > does the folder structure make sense and make it easier?
> 
> If we use `ref/sys` in the token name, like `--primer-ref-color-scale-black: #010409;` shouldn't the folder name be called the same? So it would be:
> 
> * rename `tokens/base/` to `tokens/reference/`
> * rename `tokens/functional/` to `tokens/system/`

Fair enough, I felt like I could avoid the work just for the discussion but I agree its confusing, will change it. And yes, you are correct, this would be the structure.

> > does the addition of comp tokens make sense?
> 
> Yes, we kinda need it for these [component tokens](https://github.com/primer/primitives/blob/main/data/colors/vars/component_light.ts). Ideally we can reduce them as much as possible, but we probably always need some tokens specific to a single component.

My thoughts exactly. And it feels cleaner to separate those. Also makes it easier to remove them if a component gets deprecated.

> > is the new sub-prefix structure better than the old? Does changing this have any side-effects?
> 
> I like `system` over `functional`. And yes, `reference` maybe makes it even more clear that these tokens shouldn't be used directly and only referenced in system tokens.
> 
👍 

> > we could also add another type like `gh` or `gui` or `prod` which would be similar to brand. The `gh` tokens would be stuff like `ansi` colors or `codemirror` styles that may not be specific to a component but are also not `sys` or functional colors (e.g. `--primer-gh-color-ansi-black`)
> 
> An alternative to `gh` as "the rest" or "it doesn't fit anywhere else" might be `eco`? As in "ecosystem". These tokens are used more rarely, but are still part of the overall ecosystem. e.g. `--primer-eco-color-ansi-black`

But would this include brand tokens as well? Or are you suggesting eco because e.g. `vscode` would use them as well?
I am not 100% sure as I feel if we would be better off being specific and doing custom vscode tokens if they are needed? Or do I misunderstand you?

> > * **updated** how prefixes work, now **ALL** tokens get the `--primer` prefix (we could also omit this completely for brevity or shorten to `pr`)
> 
> Yes, my vote goes to the shorter `pr` (or `gh`). Because the prefix will be used EVERYWHERE, I don't see the benefit of spelling it out. And token names are already quite long so having a shorter prefix helps a bit in reducing that.

Is this best or would you think we should drop a prefix all together?

simurai:
> > > we could also add another type like `gh` or `gui` or `prod` which would be similar to brand. The `gh` tokens would be stuff like `ansi` colors or `codemirror` styles that may not be specific to a component but are also not `sys` or functional colors (e.g. `--primer-gh-color-ansi-black`)
> > 
> > An alternative to `gh` as "the rest" or "it doesn't fit anywhere else" might be `eco`? As in "ecosystem". These tokens are used more rarely, but are still part of the overall ecosystem. e.g. `--primer-eco-color-ansi-black`
> 
> But would this include brand tokens as well? Or are you suggesting eco because e.g. `vscode` would use them as well? I am not 100% sure as I feel if we would be better off being specific and doing custom vscode tokens if they are needed? Or do I misunderstand you?

I was mostly thinking that we need a name for "the rest". A name for tokens that we don't know where else to put. So the `eco` (ecosystem) felt like a good alternative to the `gh` or `gui` or `prod` names. Other names that come to mind are: `misc`, `other`, `custom`.

And yes, `brand` could be part of it, but it depends how many tokens there are for brand and if brand has specific needs and wants to organize their tokens in a different way. I could also imagine they only import `ref` and `sys` tokens and create their own custom tokens in their repo for easier updating and might not even need it in `Primitives`? I think that's what they currently do.

But yes, also not against being more explicit and instead of having a single "the rest" folder, we create a new folder for anything that needs tokens and doesn't fit anywhere. The following list is just an example, but we could end up with something like:

- `ref` reference tokens
- `sys` system tokens
- `comp` for Primer components
- `dotcom` for "dotcom" components
- `brand` for brand
- `mobile` for iOS + Android
- `vscode` for the VS Code theme
- `desktop` for GitHub Desktop
- `edu` for https://education.github.com
- etc.

Maybe we could start with `ref`, `sys`, `comp` since we for sure need them and add more as we go along?

> > > * **updated** how prefixes work, now **ALL** tokens get the `--primer` prefix (we could also omit this completely for brevity or shorten to `pr`)
> > 
> > 
> > Yes, my vote goes to the shorter `pr` (or `gh`). Because the prefix will be used EVERYWHERE, I don't see the benefit of spelling it out. And token names are already quite long so having a shorter prefix helps a bit in reducing that.
> 
> Is this best or would you think we should drop a prefix all together?

I would say currently we don't REALLY need a prefix. But in case GitHub or other places that use `Primitives` ever wanted to mix or extend token libraries, then it would be nice to have. E.g. if you want to use `Primitives` and `Material Design` in the same project, it's nice they are prefixed. So that it's clear where they come from and you should be careful with mixing them too much:

- `pr.sys.color.success.fg` from Primer
- `md.sys.color.primary` from [Material Design](https://m3.material.io/styles/color/the-color-system/tokens#e26e130c-fa67-48e1-81ca-d28f6e4ed398)
- `my.color.text` from your own custom project tokens

I would say long term it might make sense to include a prefix and if it's short like `pr` or `gh`, that should be fine.

lukasoppermann:
> And yes, brand could be part of it, but it depends how many tokens there are for brand and if brand has specific needs and wants to organize their tokens in a different way. I could also imagine they only import ref and sys tokens and create their own custom tokens in their repo for easier updating and might not even need it in Primitives? I think that's what they currently do.

Yes, that is actually what I meant and that is what they do. I was aiming to make this work as is. 👍 

> I would say long term it might make sense to include a prefix and if it's short like pr or gh, that should be fine.
👍

lukasoppermann:
@simurai I updated the folders now. It is not all 100% split like it should be though. I change some files but much is still not done, e.g. most components are still in `tokens/components/light/components-light.json5`. I will do a clean rework of this in case we decide to go with this approach.

tokens/component/color/dark/avatar.json5
```diff
@@ -0,0 +1,25 @@
+{
```
mperrotti, line 1: The fragmentation and nesting makes token source feel overwhelming and complex. Specifically:
- having 1 file per component
- requiring more levels of folder nesting

I'd look for opportunities to reduce the number of files and levels of nesting.

### Example

#### Before

```
tokens/
├─ component/
│  ├─ color/
│  │  ├─ dark/
│  │  │  ├─ components-dark.json5
│  │  │  ├─ avatar.json5
│  │  │  ├─ counter.json5
│  │  │  ├─ overrides/
│  │  │  │  ├─ dark.high-contrast.json5
│  │  │  ├─ topicTag.json5
├─ reference/
│  ├─ color/
│  │  ├─ dark/
│  │  │  ├─ dark.dimmed.json5
│  │  │  ├─ dark.high-contrast.json5
│  │  │  ├─ dark.json5
│  │  │  ├─ dark.protanopia-deuteranopia.json5
│  │  │  ├─ dark.protanopia-tritanopia.json5
│  ├─ size/
│  │  ├─ size.json5
│  ├─ typography/
│  │  ├─ typography.json5
├─ system/
│  ├─ color/
│  │  ├─ dark/
│  │  │  ├─ ansi-dark.json5
│  │  │  ├─ checks-dark.json5
│  │  │  ├─ diff-dark.json5
│  │  │  ├─ marketting-dark.json5
│  │  │  ├─ misc-dark.json5
│  ├─ size/
│  │  ├─ border.json5
│  │  ├─ breakpoint.json5
│  │  ├─ size-coarse.json5
│  │  ├─ size-fine.json5
│  │  ├─ size.json5
│  │  ├─ viewport.json5
│  ├─ typography/
│  │  ├─ typography.json5
```

#### After
```
tokens/
├─ color/
│  ├─ components/
│  │  ├─ dark.json5
│  │  ├─ dark.overrides.json5
│  ├─ reference/
│  │  ├─ dark.dimmed.json5
│  │  ├─ dark.high-contrast.json5
│  │  ├─ dark.json5
│  │  ├─ dark.protanopia-deuteranopia.json5
│  │  ├─ dark.protanopia-tritanopia.json5
│  ├─ system/
│  │  ├─ dark.ansi.json5
│  │  ├─ dark.checks.json5
│  │  ├─ dark.diff.json5
│  │  ├─ dark.marketting.json5
│  │  ├─ dark.misc.json5
├─ size/
│  ├─ components.json5
│  ├─ reference/
│  │  ├─ border.json5
│  │  ├─ breakpoint.json5
│  │  ├─ dimensions.json5
│  │  ├─ dimensions-coarse.json5
│  │  ├─ dimensions-fine.json5
│  │  ├─ viewport.json5
├─ typography/
│  ├─ reference.json5
│  ├─ system.json5
```

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib"
-}
+}
```
mperrotti, line 3: We can probably revert this change.

This is non-blocking feedback.

mperrotti (commented):
> does the folder structure make sense and make it easier?

Yes except where otherwise noted in my PR feedback.

I think it would be even easier to understand if we explain the parts of each token's name. @langermank and @vdepizzol did a good job of this when they introduced the spacing and typography tokens - so they could probably provide good guidance. Style-dictionary also does a good job at this [in the docs](https://amzn.github.io/style-dictionary/#/tokens?id=design-token-structure).

---

> does the addition of comp tokens make sense?

Yes

---

> is the new sub-prefix structure better than the old?

I don't understand why we need the `primer` or `pr` prefix on each token. It makes the names long and doesn't add valuable information.

>  Does changing [the structure] have any side-effects?

The new token names will cause a breaking change for anybody using the tokens with the existing names.

langermank:
> the name ref (reference) tokens makes it more clear than base that those tokens are not really meant to be used directly.
> I think sys (system) tokens is a good name for our primitives, although we could also name them prim or func or fun 🤣.

Is there a compelling reason to change from `base` and `functional` to `reference` and `system` other than what we're seeing from [Material](https://m3.material.io/foundations/design-tokens/how-to-read-tokens#efa90c6e-92da-4af3-8031-f0520540df25)? I feel that the terms we came up with are already embedded into the language we regularly use to discuss tokens, and I'm not sure why we would want to make a breaking change to this now. 

> comp tokens are great to separate the overrides that are supposed to only be used by a specific component.

I'm open to including component tokens with some kind of addition to the naming convention but I think we need to address functional token names first to determine if we really need component level. 

> brand tokens could have a brand sub-prefix `(--primer-brand-color-fg-default)`

I don't totally understand why `brand` is now not considered a namespace, and everything is `primer`? This breaks the [naming convention](https://github.com/github/primer/blob/main/adrs/2022-03-31-design-token-naming-guidelines.md).

script/tempColorTokenBuild.ts
```diff
@@ -16,63 +16,89 @@ export const buildDesignTokens = (buildOptions: ConfigGeneratorOptions): void =>
   const themes: TokenBuildInput[] = [
     {
       filename: 'light',
-      source: [`tokens/functional/color/light/*.json5`],
-      include: [`tokens/base/color/light/light.json5`]
+      source: [`tokens/system/color/light/*.json5`],
+      include: [`tokens/reference/color/light/light.json5`]
```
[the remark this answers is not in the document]
langermank: I agree with @mperrotti and left a more detailed question in another comment.

tokens/component/color/dark/components-dark.json5
```diff
@@ -0,0 +1,512 @@
+{
```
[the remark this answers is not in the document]
lukasoppermann: Yes, this is the old file. I have not moved all components out yet. The idea would be to move every component into a separate file. I just didn't want to do it before we agree that this is what we want to do.

So in short: `components-dark.json5` and `components-light.json5` would not be there in the final version and be replaced with a file per component.

.vscode/settings.json
lukasoppermann: 🤣 Yes. I will not pere the PR (hence the title). If we think this is an approach we want to go with I would create a new clean PR.

tokens/component/color/dark/avatar.json5
lukasoppermann: Hey @mperrotti,
thank you for the detailed reply. 🙏 

I agree with some ideas and am unsure about others, so I hope this comment will be understandable. 😉 

>  Specifically:
>- having 1 file per component
I find this to be helpful as it makes it much easier to see which components have custom components. However I agree that this adds a lot of files that can be overwhelming.

My main concern is that if we have components that have overrides for color modes this could be a horror scenario of one property files. 👻 

> - requiring more levels of folder nesting
I think it is good to visualize it, but in the end I do agree that having a reduced folder set is nicer.

```
tokens/
├─ color/
│  ├─ components/
│  │  ├─ dark.json5
│  │  ├─ dark.overrides.json5
│  ├─ reference/
│  │  ├─ dark.dimmed.json5
│  │  ├─ dark.high-contrast.json5
│  │  ├─ dark.json5
│  │  ├─ dark.protanopia-deuteranopia.json5
│  │  ├─ dark.protanopia-tritanopia.json5
│  ├─ system/
│  │  ├─ dark.ansi.json5
│  │  ├─ dark.checks.json5
│  │  ├─ dark.diff.json5
│  │  ├─ dark.marketting.json5
│  │  ├─ dark.misc.json5
├─ size/
│  ├─ components.json5
│  ├─ reference/
│  │  ├─ border.json5
│  │  ├─ breakpoint.json5
│  │  ├─ dimensions.json5
│  │  ├─ dimensions-coarse.json5
│  │  ├─ dimensions-fine.json5
│  │  ├─ viewport.json5
├─ typography/
│  ├─ reference.json5
│  ├─ system.json5
```

I do like this approach and was thinking about something like this as well. Getting rid of the `dark` and `light` folders made a lot of sense once I added the `dark.` prefix. 

The thing that I discussed with I think @simurai before and am not sure about is using a `reference/` folder in one places and a `reference.json5` file in others. I like it, but I am afraid that this may be confusing for some people.

Do we do something like this in other places?

An alternative could be something like this:

```
├─ size/
│  ├─ components/
│  │ ├─ components.json5
│  ├─ reference/
```

lukasoppermann:
> Is there a compelling reason to change from base and functional to reference and system other than what we're seeing from [Material](https://m3.material.io/foundations/design-tokens/how-to-read-tokens#efa90c6e-92da-4af3-8031-f0520540df25)? I feel that the terms we came up with are already embedded into the language we regularly use to discuss tokens, and I'm not sure why we would want to make a breaking change to this now.

Well, this was just mentioned by @simurai and I wanted to open it up to a wider crowd for discussion. I don't feel there is a definite need to adopt this. I personally like the following about it:

- `references` to me sound much more like `don't use`. And we do always say want to stop people using those tokens.
- `functional` and `component` tokens seem, to have a large overlap, as in component tokens are component specific functional tokens. This is why separating those feels nice to me. `system` fits for me personally because they are the core tokens of the design system.

But as I said, I am happy either way, just wanted to open up the discussion.

> I'm open to including component tokens with some kind of addition to the naming convention but I think we need to address functional token names first to determine if we really need component level.

If we can completely remove component tokens that would be perfect. I am just hearing from all sides that people feel this is unlikely. But if we don't need them. This folder could just be removed.

> I don't totally understand why brand is now not considered a namespace, and everything is primer? This breaks the [naming convention](https://github.com/github/primer/blob/main/adrs/2022-03-31-design-token-naming-guidelines.md).

This is just a thought experiment extending the discussion around prefixes:
If everything is part of primer (similar to how emily sets up the docs) than brand is to. `brand` is still a namespace, just like `ref` or `sys` (or `base` and `product` or whatever), it just all has a `--primer` or `--pr` prefix.

```
[ds prefix]-[namespace]-[item]-[variant]-[property]-[modifier]
```

However I am also wondering if the prefix is necessary. I would side with @mperrotti and say we don't need it. It is just a help to play nice with other systems.

lukasoppermann:
@mperrotti 

> is the new sub-prefix structure better than the old?
> 
> > I don't understand why we need the primer or pr prefix on each token. It makes the names long and doesn't add valuable information.

This is an open question. I also feel we can probably leave it out. It just makes sure we don't have naming conflicts with other systems if they are ever used together. However I don't think this is so likely.

> Does changing [the structure] have any side-effects?
> 
>> The new token names will cause a breaking change for anybody using the tokens with the existing names.
Yes, this is going to happen either way, as we are planning on renaming tokens one way or another. But this of course is a very huge side-effect.

simurai:
> The new token names will cause a breaking change

I think that's a reason why this whole discussion kept growing. If we already make a few breaking changes, we _could_ make even more breaking changes. 😆 As in, we don't get this chance that often. On the other hand, to keep refactoring costs as low as possible, we could try to only make changes to the output if we really need to.

> I don't totally understand why `brand` is now not considered a namespace, and everything is `primer`? This breaks the [naming convention](https://github.com/github/primer/blob/main/adrs/2022-03-31-design-token-naming-guidelines.md).

Yeah, if we feel strongly about any of the ideas discussed here we should move them into these ADR PRs for final approval: 

- https://github.com/github/primer/pull/1422
- https://github.com/github/primer/pull/1309

langermank:
Hey @lukasoppermann! @vdepizzol and I brought this to today's Patterns working group to chat through it a bit more. After discussing the changes and comparing to the existing naming convention, we felt that it might be best to focus on adopting the existing naming convention for all color tokens _before_ making breaking changes to it. We may find that the existing convention works well, and if we do find issues with it we feel that it would be easier to make those changes later once all our tokens follow the same convention. I don't feel that we've had a chance to validate that the existing naming convention won't work for color.

lukasoppermann:
> Hey @lukasoppermann! @vdepizzol and I brought this to today's Patterns working group to chat through it a bit more. After discussing the changes and comparing to the existing naming convention, we felt that it might be best to focus on adopting the existing naming convention for all color tokens _before_ making breaking changes to it. We may find that the existing convention works well, and if we do find issues with it we feel that it would be easier to make those changes later once all our tokens follow the same convention. I don't feel that we've had a chance to validate that the existing naming convention won't work for color.

Hey I am very happy to explore this now that you are back.

Just to clarify I don't really feel this discussion creates breaking changes for the naming conventions. It only provides some alternatives for the first block of the convention, the `namespace/prefix`.

162517742-380a5890-371c-47be-a7af-6976546efb2b

I already worked on some tokens to see if they fit into this general idea ☝️ and for the most part it works pretty well. The "problems" that arise are not with the prefix.

However I am happy to park this and focus on working through the color tokens now that you are back. 👍

lukasoppermann:
Superseded by the [token naming ADR](https://github.com/github/primer/blob/main/adrs/2022-03-31-design-token-naming-guidelines.md)

PR #380, opened 2022-10-28
Improving type formatter

lukasoppermann:
## Summary

Now the type formatter does not run into issues if an invalid token is provided. This can happen as tokens come from json an are not type-safe.

## List of notable changes:

- **updated** [convertPropToType](https://github.com/primer/primitives/compare/improve-type-def-formatter?expand=1#)
- **added** new test cases in [`config/formats/typescript-export-definition.test.ts`](https://github.com/primer/primitives/compare/improve-type-def-formatter?expand=1#diff-3ed1bc0bb9440ad951ea7e378564b888235bf2782387ad3c4073502a48f7d5a9)

- does the change make sense

## Contributor checklist:

- [x] All new and existing CI checks pass
- [x] Tests prove that the feature works and covers both happy and unhappy paths
- [x] Any drop in coverage, breaking changes or regressions have been documented above
- [x] All developer debugging and non-functional logging has been removed
- [x] Related issues have been referenced in the PR description

## Reviewer checklist:

### DIFF
--- a/config/formats/typescriptExportDefinition.test.ts
+++ b/config/formats/typescriptExportDefinition.test.ts
@@ -10,19 +10,6 @@ describe('Format: TypeScript definitions', () => {
           $type: 'color',
           value: '#FF0000'
         }),
-        // TODO: implement types
-        // small: getMockToken({
-        //   $type: 'dimension',
-        //   value: '10px'
-        // }),
-        // big: getMockToken({
-        //   $type: 'dimension',
-        //   value: '2rem'
-        // }),
-        // shadow: getMockToken({
-        //   $type: 'shadow',
-        //   value: 'shadow value'
-        // }),
         stringValue: getMockToken({
           value: '#FF0000'
         }),
@@ -34,10 +21,6 @@ describe('Format: TypeScript definitions', () => {
             width: '1px'
           }
         }),
-        // default: getMockToken({
-        //   value: 16,
-        //   $type: 'dimension'
-        // }),
         numberValue: getMockToken({
           value: 20
         }),
@@ -47,15 +30,7 @@ describe('Format: TypeScript definitions', () => {
       }
     }
   })
-  // @ ts-expect-error: Because complex is invalid
-  //   complex: {
-  //   top: 16,
-  //   bottom: () => 16
-  // }
-  // default: getMockToken({
-  //   value: 16,
-  //   $type: 'dimension'
-  // }),
+
   it('Formats tokens adding prefix', () => {
     const input = getMockFormatterArguments({
       dictionary,
@@ -123,23 +98,6 @@ describe('Format: TypeScript definitions', () => {
     )
     expect(typescriptExportDefinition(input)).toStrictEqual(expectedOutput)
   })
-
-  // it('throws an invalidToken error if a token has no value property', () => {
-  //   const input = getMockFormatterArguments({
-  //     dictionary: getMockDictionary({
-  //       // @ts-expect-error: Because of invalid token
-  //       tokens: {
-  //         subgroup: {
-  //           red: '#FF0000'
-  //         }
-  //       }
-  //     })
-  //   })
-
-  //   expect(() => {
-  //     typescriptExportDefinition(input)
-  //   }).toThrowError('Invalid token: #FF0000')
-  // })
 
   it('throws an invalidTokenValue error if a token has an invalid value for the defined type', () => {
     const input = getMockFormatterArguments({
--- a/config/formats/typescriptExportDefinition.test.ts
+++ b/config/formats/typescriptExportDefinition.test.ts
@@ -4,14 +4,26 @@ import {format} from 'prettier'
 
 describe('Format: TypeScript definitions', () => {
   const dictionary = getMockDictionary({
-    // @ts-expect-error: Because complex is invalid
     tokens: {
       subgroup: {
         red: getMockToken({
           $type: 'color',
           value: '#FF0000'
         }),
-        yellow: getMockToken({
+        // TODO: implement types
+        // small: getMockToken({
+        //   $type: 'dimension',
+        //   value: '10px'
+        // }),
+        // big: getMockToken({
+        //   $type: 'dimension',
+        //   value: '2rem'
+        // }),
+        // shadow: getMockToken({
+        //   $type: 'shadow',
+        //   value: 'shadow value'
+        // }),
+        stringValue: getMockToken({
           value: '#FF0000'
         }),
         border: getMockToken({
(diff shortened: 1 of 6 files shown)


rezrah (approved):
This is wonderful 👏 ... 

Does this negate or overlap with the need for schema validation? 🤔 .. I like that it's built-in to the core, rather than running validation as a side-effect.

config/formats/typescriptExportDefinition.ts
```diff
@@ -117,7 +124,7 @@ const getTypeDefinition = (
   // get token type declaration from file
   const designTokenTypes: string[] = []
   for (const type of usedTypes) {
-    designTokenTypes.push(getTokenType(`${tokenTypesPath}/${type}.d.ts`))
+    designTokenTypes.push(getTokenType(`${tokenTypesPath.replace(new RegExp(/\/$/, 'g'), '')}/${type}.d.ts`))
```
rezrah, line 127: This is getting a bit long, and difficult to read. 

Can you break it up into multiple lines? And perhaps pull the regex out and either add annotations around it or write some tests for it so it's clear what it's doing?

config/formats/typescriptExportDefinition.ts
```diff
@@ -32,53 +33,59 @@ const getTokenType = (tokenTypesPath: string): string => {
...
 /**
  * convertPropToType
  * @description return take type attribute and return correct typescript type as string
  * @param type
  * @returns
  */
-const convertPropToType = (value: unknown, type?: string): string => {
-  switch (type) {
+const convertPropToType = (token: w3cTransformedToken): string => {
+  if (!Object.prototype.hasOwnProperty.call(token, 'value')) {
```
rezrah, line 54: Have you considered using `hasOwn` instead here? It's easier to read and mitigates object prototype inheritance problems. [Link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).


```suggestion
  if (!Object.hasOwn(token, 'value')) {
```

lukasoppermann:
> Does this negate or overlap with the need for schema validation? 🤔 .. I like that it's built-in to the core, rather than running validation as a side-effect.

It does not negate the need for a schema I think. But maybe it does if you have an idea of how we could dop this.
It creates a typescript type from the tokens. We could of course use it, if this somehow works.

config/formats/typescriptExportDefinition.ts
lukasoppermann: Yes, I had some ts issues some time ago, but I will give it another try

lukasoppermann: I get this:
Screenshot 2022-11-22 at 12 27 08

Even though my `tsconfig.json` has this:
```json
    "lib": ["ES2019.Array", "ES2022"],
  ```

File: config/formats/typescriptExportDefinition.ts (hunk 1 above)
Comment (lukasoppermann) [reply config/formats/typescriptExportDefinition.ts]: This ended up turning into quite a large refactor 🤣 

I refactored `toType` into a more generic `treeWalker` (now imported from another file) and removed the usedTypes as a side-effect. Instead I now run the treeWalker twice, once to build the object tree an once to get the used types.

I feel performance is not as important as clarity in this instance.

Comment (lukasoppermann) [reply config/formats/typescriptExportDefinition.ts]: I also now don't throw anymore as the tree walker ignores invalid items. It felt this was the correct approach, but let me know if you think it should throw (and why).

File: config/formats/typescriptExportDefinition.test.ts [hunk 3]
```diff
@@ -92,17 +112,53 @@ describe('Format: TypeScript definitions', () => {
...
+            booleanValue: "boolean";
           };
         };
       };`,
       {parser: 'typescript', printWidth: 500}
     )
     expect(typescriptExportDefinition(input)).toStrictEqual(expectedOutput)
   })
+
+  // it('throws an invalidToken error if a token has no value property', () => {
```
rezrah, line 127: should this be disabled?

config/formats/typescriptExportDefinition.ts
rezrah: oh weird, should be in the ES2022 spec.. not sure what issue without getting hands-on with it, so can leave as-is for now

config/utilities/treeWalker.ts
```diff
@@ -0,0 +1,31 @@
...
+    return callback(tree)
+  }
+  // is non-object value -> ignore
+  if (typeof tree !== 'object') return
+  // is obj -> continue
+  const nextObj = {}
+  for (const [prop, value] of Object.entries(tree) as [prop: string, value: unknown][]) {
+    // TODO: FIX typescript issues
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
rezrah, line 27: any chance these errors can be fixed before it merges? is the error here that the compiler can't infer types on the index below?

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "jsx": "react",
     "target": "ES2016",
     "module": "commonjs",
-    "lib": ["ES2019.Array", "ES2019.Object"],
+    "lib": ["ES2019.Array", "ES2022.Object"],
```
rezrah, line 6: Curious if we can just change this to `ESNext`? Does it still compile okay?

config/utilities/treeWalker.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * @description converts tokens to type and returns array with used type
+ * @param tree object
+ * @param callback function to be run on valid items
+ * @param validItem function to test if a node is a validItem to be transformed
+ * @returns object
+ */
+export const treeWalker = (
+  tree: Record<string, unknown>,
+  // TODO: FIX typescript issues
```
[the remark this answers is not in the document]
lukasoppermann: I agree. I also think we should have a test. If you think it is good like this from the setup I am happy to start a test and see if this makes sense.

config/formats/typescriptExportDefinition.test.ts
lukasoppermann: Yes, I removed the part as it does not throw an error anymore. Items without a `value` property get ignored now. I wrote this down somewhere in slack, but i should have rather added it here. :D

tsconfig.json
rezrah: Yeah, rather than maintaining individual specifications, we just use one catch-all.

rezrah (approved):
pre-emptively approving.. 

- we discussed fixing the type errors after this PR merges to move forward.. made a note of this separately
- removal of redundant comments / old code

🚀

PR #391, opened 2022-11-15
Border transformation

lukasoppermann:
## Summary

Adds transformer and filter to transform borders to css string.
This fixes https://github.com/github/primer/issues/1512

## Todo
- [x] add outline offset
- [x] use `color.focus`
- [x] add outline as top level to get `--ui-outline-focus` and  `--ui-outline-focus-offset`

## List of notable changes:

- **added** `config/filters/isBorder.ts`
- **added** `config/transformers/borderToCss.ts`

- does it work?
- is it okay to add borders to theme output file (they use colors so they must be tied to color modes)

## Steps to test:

1. Clone branch
1. run `npm run build:color-tokens`

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/tokens/functional/border/light.json5
+++ b/tokens/functional/border/light.json5
@@ -1,0 +1,12 @@
+{
+  outline: {
+    focus: {
+      $value: {
+        color: '{color.focus}',
+        style: 'solid',
+        width: '1px'
+      },
+      $type: 'border'
+    }
+  }
+}
--- a/config/PrimerStyleDictionary.ts
+++ b/config/PrimerStyleDictionary.ts
@@ -16,6 +16,7 @@ import {
   jsonNestedPrefixed,
   cssThemed
 } from './formats'
+import {borderToCss} from './transformers/borderToCss'
 
 /**
  * Parsers
@@ -89,6 +90,11 @@ StyleDictionary.registerTransform({
 StyleDictionary.registerTransform({
   name: 'shadow/css',
   ...shadowToCss
+})
+
+StyleDictionary.registerTransform({
+  name: 'border/css',
+  ...borderToCss
 })
 
 /**
--- a/config/filters/isBorder.test.ts
+++ b/config/filters/isBorder.test.ts
@@ -1,0 +1,18 @@
+import {getMockToken} from '~/src/test-utilities'
+import {isBorder} from './isBorder'
+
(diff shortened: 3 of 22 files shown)


tokens/functional/border/light.json5
```diff
@@ -0,0 +1,12 @@
+{
```
langermank, line 1: This isn't a blocking comment, but for focus styles specifically we use `outline` not `border` for most cases. I probably wasn't very clear about that, sorry! We also are somewhat flexible about the value as we tweak it for specific cases: https://github.com/primer/css/blob/main/src/support/mixins/misc.scss#L29

I think this PR is still needed though for other border tokens?

langermank: Just saw in the issue you mention outline is similar enough to border.. that does make sense, but we'll also want tokens for `outline-offset` which unfortunately is not part of the outline shorthand. So just something to think about (we probably want all our focus style tokens to have the same property in the name)

lukasoppermann: Ahh, I did not know about `outline-offset`. However it seems [the offset](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset) is not part of the [outline shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/outline).

So I would suggest the offset should be defined as a dimension/size token.

lukasoppermann: Changed the top level name to `outline` and added the `offset`. However it is currently not being complied because we have not moved the sizes, etc. over to the new build system.

Another question that is related to the naming is if the offset should be:
- `outline-focus-offset`
- `offset-outline-focus`
- `size-outline-offset-focus`
- ...

But this can be fixed after this is merged.

langermank: Yeah, we can fix this later once we allow camelCase in the new convention. I think it would be:

```
ui-color-outline-focus 
ui-size-outline-focus
ui-size-outlineOffset-focus
```

lukasoppermann: We would also need something like:
```sh
--ui-border-outline-focus: [color] [style] [width];
# or 
--ui-outline-focus: [color] [style] [width];
```

langermank (approved):

PR #400, opened 2022-11-23
Add prefix to name transformers

lukasoppermann:
## Summary

Incorporate prefix like `product` or `brand` into name transformers so that they are added to output like `.css` variables or in the deprecated json files.

## List of notable changes:

- **added** tests to test if prefix is added
- **updated** `namePathToDotNotation.ts` and `namePathToKebabCase.ts` to add prefix if defined

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/config/transformers/namePathToKebabCase.ts
+++ b/config/transformers/namePathToKebabCase.ts
@@ -8,5 +8,12 @@ import StyleDictionary from 'style-dictionary'
  */
 export const namePathToKebabCase: StyleDictionary.Transform = {
   type: `name`,
-  transformer: (token: StyleDictionary.TransformedToken) => token.path.join('-')
+  transformer: (token: StyleDictionary.TransformedToken, options?: StyleDictionary.Platform): string => {
+    return (
+      [options?.prefix, ...token.path]
+        // remove undefined if exists
+        .filter((part: unknown): part is string => typeof part === 'string')
+        .join('-')
+    )
+  }
 }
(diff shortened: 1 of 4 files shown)


config/transformers/namePathToKebabCase.ts
```diff
@@ -8,5 +8,12 @@ import StyleDictionary from 'style-dictionary'
  */
 export const namePathToKebabCase: StyleDictionary.Transform = {
   type: `name`,
-  transformer: (token: StyleDictionary.TransformedToken) => token.path.join('-')
+  transformer: (token: StyleDictionary.TransformedToken, options?: StyleDictionary.Platform): string => {
+    return (
+      [options?.prefix, ...token.path]
+        // remove undefined if exists
+        .filter((part: unknown): part is string => typeof part === 'string')
```
rezrah, line 15: do you need to narrow type with `is` here? what happens if you just put part:string? or better yet, infer type from options?

rezrah (approved):

config/transformers/namePathToKebabCase.ts
lukasoppermann: I do, yes. Otherwise I get a ts error as ts does not understand that the filter will only allow strings to pass through.
If options is not defined the array has an undefined value in it, right?

I had tried `<string | undefined>[options?.prefix, ...token.path]` and even `<string>[options?.prefix, ...token.path]` but both gave me errors.

Is there a problem with `part is string` in general?

rezrah: not a problem with it, I was just reading it and confused why we needed unknown, which feels a bit icky.. and assumed type inference would just work 🤷‍♂️ . Carry on!

PR #401, opened 2022-11-23
Add postinstall script to install deps storybook

lukasoppermann:
## Summary

Automatically install storybook deps on `npm i`

## List of notable changes:

- **added** new npm scripts `install:storybook` and `postinstall`
- **updated** package.json to have scripts in alphabetical order

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,9 +27,12 @@
   "homepage": "https://github.com/primer/primitives#readme",
   "scripts": {
     "build": "ts-node ./script/build.ts && tsc --project tsconfig.build.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,12 +41,11 @@
     "prebuild:color-tokens": "rm -rf tokens-v2-private/**/**/color",
     "prepack": "yarn build && yarn build:tokens",
     "prepare": "husky install",
+    "postinstall": "npm run install:storybook",
```
rezrah, line 44: eek.. this will also install for users of the npm package. 10/10 don't recommend. 

can you put this into the prepare command instead?

lukasoppermann: Ahhh, okay, I didn't know `prepare` runs on local npm install as well. 👍  will change it.

PR #424, opened 2022-12-01
moved utilts and @types out of src

lukasoppermann:
## Summary

Moved `utils.ts` and `@types` out of `src`

- does it still work without breaking

## Steps to test:

<!--

2. Go to # page

1.
1.
1.

-

## Contributor checklist:

## Reviewer checklist:


rezrah:
> Moved utils.ts out of src

What was the reason for moving this out of `src`? I'm not sure I would expect to find these in 'scripts'. In the long term, I'd expect src to contain anything with contributes towards output, and scripts to contain the files that triggers the execution of the output.

lukasoppermann:
> What was the reason for moving this out of src? 
This will be removed once we move to the new build script.

> I'm not sure I would expect to find these in 'scripts'. In the long term, I'd expect src to contain anything with contributes towards output, and scripts to contain the files that triggers the execution of the output.

I am not sure I follow why. The idea was to move tokens into `src` as the `json` files are the source for the output with are the tokens. 

I can also see an argument for having the build flow in source. What I currently don't understand is why we would want to have the imported files for scripts in src? Can you elaborate on this?

I could see us moving everything into source including config and maybe adding a `build.ts` in the root. Or just move everything into source. 

I could also see tokens staying outside of src in a tokens folder.

Just to clarify, I am not saying we can't do it. I just don't understand why this specific split makes sense.

rezrah:
> I am not sure I follow why. The idea was to move tokens into src as the json files are the source for the output with are the tokens.

Moving token source files into `src` seems sensible, as they're precompiled. Agree with that. The file you're moving _out_ of `src` seems to be [this one](https://github.com/primer/primitives/blob/80885498dc30a3103d96ad1b7f5ee07248b931ab/script/lib/utils.ts). IMO, as this is just another required input to achieve the output, I'd expect this to stay in `src`. It's also a `.ts` file, so arguably  should be transpiled if you wanted to call it outside of `src` if we wanted to be purist about it. 

> I can also see an argument for having the build flow in source. What I currently don't understand is why we would want to have the imported files for scripts in src? Can you elaborate on this?

Yeah, I would expect the build flow source code to live in `src/build/` (or similar) . The execution scripts, whether it's written in TS or Shell to live in `scripts/`, which calls either a compiled version of src build scripts, or a pre-compiled entrypoint  in the `src/build`.

something like 

```js
// scripts/build.ts

import {build} from "~src/build/index.ts"

build(stuff);
```

> I could also see tokens staying outside of src in a tokens folder.

I guess you could, but I personally see that living in `src/tokens`.

I think we should declutter the root and make it really standardised. 

- `src/` => contributes to output..
- `scripts/` => triggers execution of output using src
- `dist/` => is the output

Right now it feels like we have an assortment of source files in different folders, so someone coming in new to the repo may not realise that something is source and therefore requires particular care in. handling.

lukasoppermann:
> I think we should declutter the root and make it really standardised.
> 
> src/ => contributes to output..
scripts/ => triggers execution of output using src
dist/ => is the output

Okay, this makes sense to me, I agree with this. We should also rename `script` to `scripts`, right? This always trips me up.

One question, the `config` folder. Where to we move this? I think it would belong into the bild folder. Maybe even without the config around it?

rezrah:
> We should also rename script to scripts, right? This always trips me up.

😅 yes, 💯 

> One question, the config folder. Where to we move this? I think it would belong into the bild folder. Maybe even without the config around it?

that works for me.. i can't remember exactly what's in there, but my approach there would be to put shared configs in the root of `src`.. and any config specific to build-only in `src/build/conf`

PR #426, opened 2022-12-01
refactor copyFiles function to remove fs-extra

lukasoppermann:
## Summary

Refactor `copyFilesAndFolders` to `copyFromDir`.

- It now only works with on source and one desitionation
- It can only copy the content of an entire folder

**Reason:** This is the only use case we have.

## Contributor checklist:

## Reviewer checklist:


.eslintrc.js
```diff
@@ -88,6 +88,7 @@ module.exports = {
       extends: ['plugin:@typescript-eslint/recommended', 'prettier'],
       rules: {
         'no-shadow': 'off',
+        'import/no-nodejs-modules': 'off',
```
rezrah, line 91: Guessing you're wanting to turn this off for server-side scripts? If so, I'd recommend creating a more targetted override set. You've added this one into `['**/*.{ts,tsx}']`, which includes files that can be run client-side, where node modules shouldn't be run.

lukasoppermann: Okay, sounds good. How do I do this?

Should I just do this for `['config/**/*.ts','script/**/*.ts']`?

config/utilities/copyFromDir.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @param destination path
+ * @returns promise
+ */
+export const copyFromDir = async (source: string, destination: string) => {
+  // adjust trailing slash
+  const dest = `${destination.replace(/\/$/, '')}/`
+  const src = `${source.replace(/\/$/, '')}`
+  execSync(`mkdir -p ${dest}`)
+  return exec(`cp -R ${src}/* ${dest}`)
+}
```
rezrah, line 16: just curious if you'd considered these alternatives...

a) using `fs` as it has better process parallelisation 
b) running natively through a shell script, to avoiding blocking the event loop and saturating the JS thread with something like `execSync`

.eslintrc.js
rezrah: yeah that sounds good for now, and can be adjusted later if needed 👍

config/utilities/copyFromDir.ts
lukasoppermann: No, neither of those.

I would like to keep it in the script, which is why I think **b)** is not so good. 

I am happy to do **a)** if you can point me to where I can learn how to. This is just the way I know. 🙃

rezrah: to expand on a) I think you can pass `recursive` option to `fs.cp`, which can achieve the same thing. I'm totally onboard with dropping fs-extra for something like this though.

lukasoppermann: Wait, you mean:
- fs.readdirSync
- loop → fs.copyFile?

lukasoppermann: `fs.cp` is experimental though, is this not a problem?

rezrah: either of the approaches you mentioned would be fine. We're at node 18 now, so not long to go to 19 where fs.cp comes in. Happy to play it safe though and use fs.copyFile instead.

lukasoppermann: If we go with `fs.cp` do we need to add node@18 to our package.json? We currently don't have node in there. But we should maybe add it at least as a peer-dep? 🤔

rezrah: Yeah it'll run the background, but the problem there may be that you'll run into race conditions if your function returns too early. If it works, best to leave as-is. I think it's fine. Alternatively, add the await or use `copyFileSync` maybe?

rezrah (approved):
Thanks for following up on removing fs-extra 🙇 ... latest updates look great 🤩

config/utilities/copyFromDir.ts
```diff
@@ -0,0 +1,21 @@
+/* istanbul ignore file */
+import {copyFile, readdir, mkdir} from 'fs/promises'
+/**
+ * copyFromDir
+ * @description Copies all files from source folrder to destination
```
rezrah, line 5:
```suggestion
 * @description Copies all files from source folder to destination
```

PR #427, opened 2022-12-02
move v3 tokens to src/tokens

lukasoppermann:
## Summary

This PR copies/moves token source files to `src/tokens` to be changed without affecting v2 tokens.
It also changes the output dest for the new build flow to `tokens-v3-private/`

- does everything compile fine?
- is something missing?

## Contributor checklist:

## Reviewer checklist:


script/tempColorTokenBuild.ts
```diff
@@ -5,95 +5,105 @@ import {typeDefinitions, deprecatedJson, css, docJson, scss, javascript, typescr
...
+
 export const buildDesignTokens = (buildOptions: ConfigGeneratorOptions): void => {
   const themes: TokenBuildInput[] = [
     {
       filename: 'light',
       source: [
-        `tokens/functional/color/light/*.json5`,
-        `tokens/functional/shadow/light.json5`,
-        `tokens/functional/border/light.json5`
+        `${dirTokensFunctional}/color/light/*.json5`,
```
rezrah, line 17: I can totally see why you did this, but I think it's an optimisation at the expense of readability. I think I preferred the explicit paths as I didn't have to think at all about what it resolves to. Not blocking, but thought I'd share anyway.

.eslintignore
```diff
@@ -1,4 +1,5 @@
 tokens-v2-private/
+tokens-v3-private/
```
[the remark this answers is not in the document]
lukasoppermann: I think the new ones we will release would be v3, yes. But this is currently to have a safe experimentation / build field. I think the CI stuff is only needed once we actually want to release it, even if experimentally, right?

script/tempColorTokenBuild.ts
lukasoppermann: Hmm, I guess it may be harder to read. I am fine reverting this.

.eslintignore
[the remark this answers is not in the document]
rezrah: I would recommend you have it even for experimentation. I find it's most useful for maintainers like yourself and @langermank who are regularly adding and updating tokens quickly. Your call though, happy to leave it for now too.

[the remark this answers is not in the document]
lukasoppermann: Oh, sure, if you can quickly add it, that would be great.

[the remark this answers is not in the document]
lukasoppermann: No, I was not planning on shipping them yet.

scripts/tempColorTokenBuild.ts
```diff
@@ -193,6 +199,6 @@ export const buildDesignTokens = (buildOptions: ConfigGeneratorOptions): void =>
 }
 
 buildDesignTokens({
-  buildPath: 'tokens-v2-private/',
+  buildPath: 'tokens-v3-private/',
```
rezrah, line 202: So this means that we're no longer supporting / issuing future updates to tokens in `tokens-v2-private/`. Wondering if we need to communicate that somewhere 🤔

lukasoppermann: No, because the script we use for this is unchanged. So we can still issue updates.

rezrah (approved):
Thanks for rolling back the change to explicit paths,  reads a lot clearer to me ☺️.

There's a merge conflict so can't confirm the deployments worked, but assuming they are.. happy to go ahead.

scripts/tempColorTokenBuild.ts
lukasoppermann: This is the build script for tokens that are currently shipped https://github.com/primer/primitives/blob/main/build.js

PR #485, opened 2023-02-08
add control border emphasis

lukasoppermann:
## Summary

Adding a new border that has a `3:1` contrast ratio against the background.

@langermank the token name does not work 100%. It will now be: `--color-control-border-color-emphasis` because this is how the old script converts the name.

I set it up as this in json:

```json
{
  "control": {
    "borderColor": {
      "emphasis": "#858F99"
    }
  }
}
```

## List of notable changes:

- **added** `control.borderColor.emphasis`

### DIFF
--- a/.changeset/clean-weeks-taste.md
+++ b/.changeset/clean-weeks-taste.md
@@ -1,0 +1,5 @@
+---
+'@primer/primitives': minor
+---
+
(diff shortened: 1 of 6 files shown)


langermank (approved):
Looks great! No worries about `color-`, we'll address it with all the other find/replace 😄

PR #490, opened 2023-02-10
Adding swift colorset action

lukasoppermann:
This PR adds an action that creates colorsets for swift from the provided color tokens.


scripts/buildiOS.ts
```diff
@@ -0,0 +1,76 @@
...
+    }).buildAllPlatforms()
+    // build base scales
+  }
+}
+
+/** -----------------------------------
+ * Run build script
+ * ----------------------------------- */
+buildDesignTokens({
+  buildPath: 'tokens-v3-private/',
```
jonrohan, line 74: You might need to add this path to the files config so it's included in the npm package? https://github.com/primer/primitives/blob/iOS-colorsets/package.json#L5-L10

lukasoppermann: Thanks. At the moment it is in demo mode and not meant to be available.  I am still trying to figure out in here: https://github.com/github/primer/issues/1698 how to best distribute it for iOS. You do not by any chance have iOS experience?

jonrohan: > You do not by any chance have iOS experience?

sorry, have practically zero experience

lukasoppermann: Same here. 🤣

PR #508, opened 2023-03-09
Add script to check for missing tokens

lukasoppermann:
## Summary

As discussed with @langermank this script checks if there are any tokens missing in the themes compared to the default `light` theme.

### DIFF
--- a/scripts/diffThemes.ts
+++ b/scripts/diffThemes.ts
@@ -1,0 +1,66 @@
+import {PrimerStyleDictionary} from '~/src/PrimerStyleDictionary'
+// @ts-expect-error: no types exists for style-dictionary createDictionary
+import createDictionary from 'style-dictionary/lib/utils/createDictionary'
+import {themes as themesConfigArray} from './themes.config'
+import type StyleDictionary from 'style-dictionary'
+
+PrimerStyleDictionary.registerTransform({
+  type: `value`,
+  transitive: true,
+  name: `valueToNull`,
+  transformer: _token => null,
+})
+
+const tokenNameArray = ({dictionary}: {dictionary: StyleDictionary.Dictionary}) =>
+  dictionary.allTokens.map(({name}: {name: string}) => name)
+
+const themes = themesConfigArray.map(({filename, source, include}) => {
+  const sd = PrimerStyleDictionary.extend({
+    source,
+    include,
+    platforms: {
+      json: {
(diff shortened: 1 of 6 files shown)


scripts/diffThemes.ts
```diff
@@ -0,0 +1,66 @@
+import {PrimerStyleDictionary} from '~/src/PrimerStyleDictionary'
+// @ts-expect-error: no types exists for style-dictionary createDictionary
+import createDictionary from 'style-dictionary/lib/utils/createDictionary'
+import {themes as themesConfigArray} from './themes.config'
+import type StyleDictionary from 'style-dictionary'
+
+PrimerStyleDictionary.registerTransform({
```
rezrah, line 7: Should this be here? I thought we put our transforms in a dedicated folder or is this a transient config change only needed for the diff?

rezrah (approved):
✅ to help move this forward.

Left a few suggestions / questions, but nothing blocking.

scripts/diffThemes.ts
```diff
@@ -0,0 +1,66 @@
...
+    }),
+  ]
+})
+
+/**
+ *
+ * @param mainThemeName - name of the theme (from themes.config.ts) to compare all other themes against
+ */
+
+const diffThemes = (mainThemeName: string) => {
```
rezrah, line 40: Would be cool to have some tests for this with happy/unhappy path data to confirm it works as expected.. but from scanning the code it looks good 👍

scripts/diffThemes.ts
lukasoppermann: Actually, good spotting, this is a leftover, I ended up not needing it.

It used to be for this diff only though.

scripts/diffThemes.ts
```diff
@@ -0,0 +1,66 @@
...
+  const mainIndex = themes.findIndex(([name]) => name === mainThemeName)
+  const mainTokens: string[] = themes.splice(mainIndex, 1)[0][1] as string[]
```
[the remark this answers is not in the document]
lukasoppermann: Yes, good point, with using a local copy. I don't think the index is brittle though, given that we get the index by theme name before. But I will rework this.

scripts/diffThemes.ts
lukasoppermann: Yes, I will look into this as a separate PR. 👍

PR #509, opened 2023-03-09
disable import/no-unresolved for dist files to avoid issues when committing

lukasoppermann:
## Summary
Currently if you don't run `npm run build` before commiting eslint throws an import error for the color check and the docs.

This happens even if you didn't work on this part at all. When the docs are being built `npm run build` is run by default, so the error does not help.

I disabled eslint for those 4 imports using:

```js
// eslint-disable-next-line import/no-unresolved
```

### DIFF
--- a/docs/src/components/color-theme-picker.tsx
+++ b/docs/src/components/color-theme-picker.tsx
@@ -1,6 +1,7 @@
 import {Box} from '@primer/react'
 import React from 'react'
 import {sentenceCase} from 'sentence-case'
+// eslint-disable-next-line import/no-unresolved
 import colors from '../../../dist/js/colors'
 import {useColorTheme} from './color-theme-context'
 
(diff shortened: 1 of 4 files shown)


docs/src/components/color-theme-picker.tsx
```diff
@@ -1,6 +1,7 @@
 import {Box} from '@primer/react'
 import React from 'react'
 import {sentenceCase} from 'sentence-case'
+// eslint-disable-next-line import/no-unresolved
```
rezrah, line 4: If someone is attempting to run docs, I think we should run the build as a prerequisite given it's a dependency and avoid suppressing errors. ESLint is technically correct in warning the user about this here imo.

lukasoppermann: As discussed, we do run build before storybook and docs.

rezrah (approved):
👍 

Few notes for posterity and future readers.

- the error should ideally be prevented by building the lib, which is a dependency of the docs rather than suppressing the error, which is valid
- using linter comments to suppress these presents maintainability challenges as you need to keep adding them as more `dist` imports are provided

Despite this, in the interest of not extending duration of pre-commit hooks, build times _and_ given that these docs will be deprecated soon; the sustainability issue may not be a problem in the long term

PR #510, opened 2023-03-09
add test for diffThemes

lukasoppermann:
## Summary

Tests for new diffThemes script

### DIFF
--- a/scripts/diffThemes.test.ts
+++ b/scripts/diffThemes.test.ts
@@ -1,0 +1,34 @@
+import {diffThemes} from './diffThemes'
+
+describe('diffThemes', () => {
+  const themesArray: [string, string[]][] = [
+    ['light', ['color', 'color.text']],
+    ['dark', ['color', 'color.text']],
+  ]
+
+  it('Throws error if mainTheme name does not exist', () => {
+    expect(() => diffThemes('invalid', themesArray)).toThrow()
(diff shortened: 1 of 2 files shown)


scripts/diffThemes.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    ['light', ['color', 'color.text']],
+    ['dark', ['color', 'color.text']],
+  ]
+
+  it('Throws error if mainTheme name does not exist', () => {
+    expect(() => diffThemes('invalid', themesArray)).toThrow()
+  })
+
+  it('Outputs message for missing tokens', () => {
+    const withMissing = JSON.parse(JSON.stringify(themesArray))
```
rezrah, line 14: I'm possibly missing something fairly obvious, but do you need to stringify and reparse this array? What does it accomplish?

rezrah (approved):
🚀 

Thanks for adding this ✨. I love reading tests, as they help explain exactly how things work and hopefully also help future maintainers refactor more easily. I like how you covered the unhappy paths 💖

scripts/diffThemes.test.ts
lukasoppermann: > Are you trying to avoid mutating the original array?

Yes, and since it is nested a normal spread does not do it.

PR #514, opened 2023-03-14
A11y update to fix contrast issue with primary green

lukasoppermann:
## Summary

This PR fixes the failing contrast between the white text and green button background as part of https://github.com/github/primer/issues/1648.

It addresses both the current (old) implementation as well as the new v3 tokens.

### DIFF
--- a/data/colors/vars/global_light.ts
+++ b/data/colors/vars/global_light.ts
@@ -40,7 +40,7 @@ export default {
   },
   success: {
     fg: get('scale.green.5'),
-    emphasis: get('scale.green.4'),
+    emphasis: '#1f883d',
     muted: alpha(get('scale.green.3'), 0.4),
     subtle: get('scale.green.0')
   },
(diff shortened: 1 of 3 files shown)


data/colors/vars/global_light.ts
```diff
@@ -40,7 +40,7 @@ export default {
   },
   success: {
     fg: get('scale.green.5'),
-    emphasis: get('scale.green.4'),
+    emphasis: '#1f883d',
```
simurai, line 43: Is this on purpose? Currently we use `#2c8747` in the [feature flag](https://github.com/github/github/blob/440fea807e9348e719371a0545a9b2a6cb122608/app/assets/stylesheets/bundles/color-modes-contrast/index.scss#LL13C31-L13C38).

Or is it to match the output from the new `tokens/`

```
       mix: {
          color: '{base.color.green.5}',
          weight: 0.75,
        },
```

simurai (approved):

data/colors/vars/global_light.ts
lukasoppermann: Its to match the output.

PR #529, opened 2023-03-26
Add Primitives v8 colors

langermank:
## Summary

A much cleaner version of https://github.com/primer/primitives/pull/478 😄 

This is the initial setup for testing our next major v8 release. Everything will be tested behind a feature flag to start, but ideally we avoid visual bugs by testing all of this in Storybook first! 🤞 

## Notes

- I removed some GitHub specific tokens that we have moved to dotcom, such as Checks and some old marketing tokens
- I'm keeping the shadow values the same for an easier rollout, and we can revisit later
- All states are using our new formula with mix/alpha, which I will document soon

This is a challenging PR to review. I would recommend focusing on visually testing in Storybook. Please focus on the following:

- [ ] Ensure that I have copied over all a11y color updates correctly
- [ ] Test the `Button` **variants** and **states** in all color themes. I did quite a bit of work on these. I feel like something might be off about the `primary` border → Storybook link
- [ ] Test all "control" type components in all color themes → Storybook link
- [ ] Scan through the comparison tables for large discrepancies → Storybook link

- Part of https://github.com/github/primer/issues/1345

## Contributor checklist:

## Reviewer checklist:


src/platforms/css.ts
```diff
@@ -7,10 +7,10 @@ const getCssSelectors = (outputFile: string): {selector: string; selectorLight:
   const lastSlash = outputFile.lastIndexOf('/')
   const outputBasename = outputFile.substring(lastSlash + 1, outputFile.indexOf('.'))
   const themeName = outputBasename.replace(/-/g, '_')
-  const mode = outputBasename.substring(0, 4) === 'dark' ? 'dark' : 'light'
+  // const mode = outputBasename.substring(0, 4) === 'dark' ? 'dark' : 'light'
 
   return {
-    selector: `[data-color-mode="${mode}"][data-${mode}-theme="${themeName}"]`,
+    selector: `[data-color-mode="light"][data-light-theme="${themeName}"], [data-color-mode="dark"][data-dark-theme="${themeName}"]`,
```
lukasoppermann, line 13: Why are you changing this?

simurai (commented):
The diff is kinda hard to read. So I mostly clicked through [Storybook](https://primer-5aaf7c1c32-26615723.drafts.github.io/storybook).

Haven't really noticed anything except maybe the Banners in the dark theme:

Before | After
--- | ---
Screen Shot 2023-03-27 at 18 23 20 | Screen Shot 2023-03-27 at 18 22 40

But I assume that's intended?

Screen Shot 2023-03-27 at 18 22 07

langermank:
@simurai @lukasoppermann I think I've addressed the alpha issue with status colors in dark mode 🤞

simurai:
> I think I've addressed the alpha issue with status colors in dark mode 

Yes, Banners look good. 👍 

Scanned through the [Tables > Primitives](https://primer-5aaf7c1c32-26615723.drafts.github.io/storybook/?path=/story/migration-tables--primitives&globals=theme:dark) in the dark theme again:


Replacing `color-neutral-subtle` with `bgColor-neutral-muted` would have a bigger color difference. Is that intentional? Haven't checked where we use that.

Screen Shot 2023-03-28 at 10 18 02

Also, I the `bgColor-closed-muted` seems to still use more transparency:

Screen Shot 2023-03-28 at 10 22 22

And these two, that I'm not sure 🤔 :

Screen Shot 2023-03-28 at 10 27 31
Screen Shot 2023-03-28 at 10 27 44

langermank:
@simurai thank you! Great catch, appreciate that attention to detail ✨ 

I fixed both `bgColor-neutral-muted` and `bgColor-closed-muted`. The other two I'm not really worried about. We introduced a new backdrop color with Overlay, and everyone felt it was too light. So I worked on a new value somewhere in between the new and old :D 

I don't think that border token is used much, and we have more control now over "control" borders. So I'm essentially just removing it.

simurai (approved):
> I fixed both `bgColor-neutral-muted` and `bgColor-closed-muted`.

👍 Ok, maybe good to ship? 😅 

Doesn't mean everything needs to be super final and we can make follow-ups if we discover problems once we start to migrate to the new variables.

src/platforms/css.ts
lukasoppermann: If this is really the case, I will do a PR just for this, so that it is not hidden in this giant one and can be more easily tracked.

However I am not sure this is actually correct. Maybe @colebemis can add to this discussion.
In the current version we get the following:

```css
[data-color-mode="light"][data-light-theme="light_colorblind"],
[data-color-mode="auto"][data-light-theme="light_colorblind"] {

or

[data-color-mode="dark"][data-dark-theme="dark_high_contrast"],
[data-color-mode="auto"][data-light-theme="dark_high_contrast"] {
```
With the proposed changes we get:

```css
[data-color-mode="light"][data-light-theme="light_colorblind"],
[data-color-mode="dark"][data-dark-theme="light_colorblind"],
[data-color-mode="auto"][data-light-theme="light_colorblind"] {

or

[data-color-mode="light"][data-light-theme="dark_high_contrast"],
[data-color-mode="dark"][data-dark-theme="dark_high_contrast"],
[data-color-mode="auto"][data-light-theme="dark_high_contrast"] {
```

What is added is a line I think will never execute. As far as I see you can't get a combination of `data-color-mode="dark"` and a light theme like `data-dark-theme="light_colorblind"`. The only way you can get a light theme in _dark mode_ is using `auto`.

Am I missing something?

jonrohan: > What is added is a line I think will never execute. As far as I see you can't get a combination of `data-color-mode="dark"` and a light theme like `data-dark-theme="light_colorblind"`. The only way you can get a light theme in _dark mode_ is using `auto`.
> 
> Am I missing something?

Yeah that's correct, in dotcom ui, when selecting single mode, you'll only get the corresponding `data-color-mode` to the stored theme value. The mismatch only happens if someone selects the opposite theme for there light/dark auto themes.

langermank: Removing this change for now, and we can bring it back later if we find it necessary in dotcom!

PR #531, opened 2023-03-28
rename v3 to next and add to export

lukasoppermann:
## Summary

- **updated** `v3` to `next`
- **added** `tokens-next-private` to `dist` folder


simurai (approved):
What's the difference to v2? The new colors?

And is there a chance we need another version after `next`? Before we can clean it all up? `next-v2`? 😬 

If ☝️ is no concern, then should be fine. 👍

PR #546, opened 2023-03-29
Add customMedia format

lukasoppermann:
## Summary

This PR adds a new format for css `@custom-media`.

It is used on any token of the type `custom-viewportRange`

It is currently only used in css.


langermank (approved):
Thank you!

PR #555, opened 2023-04-04
Fix for primary color in cvd modes

lukasoppermann:
## Summary

Adds an exception for primary to allow color modes to keep the correct color.

### DIFF
--- a/.changeset/polite-penguins-wonder.md
+++ b/.changeset/polite-penguins-wonder.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Fix for primary color in cvd modes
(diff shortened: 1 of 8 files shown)


langermank (approved):

simurai (approved):
If I understood the changes correctly:

- the change from black: `#F10409`, to black: `#010409` was just a typo
- adding all the exceptions for `accent-fg` to `blue.3` is needed since we changed the "base/default" dark theme and all the other themes inherit the change
- `success.emphasis` + `danger.fg` in color blind is needed because it uses a hardcoded value in the "base/default" theme

Ok, yes.. I think this all makes sense. 👍 I can merge and create a new release and add it to dotcom.

PR #585, opened 2023-04-20
consumer_test-onChange

lukasoppermann:
## Summary

Only run consumer test when files change that actually effect the output.

### DIFF
--- a/.github/workflows/consumer_test.yml
+++ b/.github/workflows/consumer_test.yml
@@ -10,8 +10,29 @@ env:
   TEST_FOLDER: primer-brand
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    # Required permissions
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      outputEffected: ${{ steps.filter.outputs.outputEffected }}
+    steps:
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        list-files: shell
+        filters: |
+          outputEffected:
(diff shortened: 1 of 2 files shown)


.github/workflows/consumer_test.yml
```diff
@@ -10,8 +10,29 @@ env:
...
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    # Required permissions
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      outputEffected: ${{ steps.filter.outputs.outputEffected }}
```
rezrah, line 20: Nit, but should this be `outputAffected` to use past participle instead of effect? 

```suggestion
      outputAffected: ${{ steps.filter.outputs.outputAffected }}
```

lukasoppermann: Good catch, thanks

.github/workflows/consumer_test.yml
```diff
@@ -10,8 +10,29 @@ env:
...
+    steps:
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v2
```
rezrah, line 23: I've not used this workflow before, but it seems reasonably well maintained so fine to go with this for now 👍... 

Just curious if you'd considered running `git diff` to do this, or use `actions/github-script` to do it more programmatically.

E.g. 

```suggestion
    - name: Check for file changes
       run: |
          if ! git diff --name-only main -- .; then
            echo "::set-output name=filesChanged::true"
          fi
```

☝️ if git diff finds a difference between this branch and main it will change the output of the step to true

lukasoppermann: Hmm, I think I will merge this and look into using the rest api https://octokit.github.io/rest.js/v19#pulls-list-files

PR #609, opened 2023-04-26
Add overview table for typography docs

langermank:
- Overview table in Storybook
- Add descriptions to shorthand tokens

### DIFF
--- a/src/tokens/functional/typography/typography.json
+++ b/src/tokens/functional/typography/typography.json
@@ -37,7 +37,8 @@
           "lineHeight": "{text.display.lineHeight}",
           "fontFamily": "{fontStack.sansSerif}"
         },
-        "$type": "typography"
+        "$type": "typography",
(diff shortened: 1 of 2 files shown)


src/tokens/functional/typography/typography.json
```diff
@@ -37,7 +37,8 @@
           "lineHeight": "{text.display.lineHeight}",
           "fontFamily": "{fontStack.sansSerif}"
         },
-        "$type": "typography"
+        "$type": "typography",
+        "$description": "Hero-style text for brand to product transition pages. Utilize Title (large) styles on narrow viewports."
```
langermank, line 41: @lukasoppermann is there a way to add something like this to JSON but not output it as a CSS comment? I just want access to it in JSON docs 🤔

langermank: I think that would be good. No one is looking at these CSS files really (I dont think)

PR #623, opened 2023-04-27
test(e2e): add playwright setup

joshblack:
Playwright snapshot tests help us track visual regressions by storing screenshots of all Storybook stories. This implementation will capture all stories, but we can always pair it down in the future if it seems like too much.

- Add Playwright
- Configure snapshot tests action to run on all PRs
- Configure `update snapshots` label


.github/workflows/ci.yml
```diff
@@ -97,3 +97,54 @@ jobs:
 
       - name: Run unit tests
         run: npm test
+
+  vrt-runner:
```
lukasoppermann, line 101: Can we add a name here that is more descriptive? So that it does not say `CI / vrt`

.github/workflows/ci.yml
```diff
@@ -97,3 +97,54 @@ jobs:
...
+        run: kill ${{ steps.storybook.outputs.pid }}
+      - name: Upload report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: vrt-${{ matrix.shard }}
+          path: .playwright/report
+
+  vrt:
+    if: ${{ always() }}
```
lukasoppermann, line 144: What does always() do? Does it really run always, if so can't we just remove the if?

.github/workflows/ci.yml
```diff
@@ -97,3 +97,54 @@ jobs:
...
+      - name: Stop storybook
+        run: kill ${{ steps.storybook.outputs.pid }}
+      - name: Upload report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: vrt-${{ matrix.shard }}
+          path: .playwright/report
+
+  vrt:
```
lukasoppermann, line 143: Can we add a name here that is more descriptive? So that it does not say `CI / vrt-runner (1)`

lukasoppermann (changes requested):

joshblack:
Hey @lukasoppermann! 👋 Just was putting this up as a draft from the pairing session @langermank and I had last week 👀 Let me know if it'd be preferred to leave it open as a draft or close it out.

.github/workflows/ci.yml
joshblack: One of the challenges with the "runner" pattern is that we need a way to report the results of these runners in a single required status check (mostly to avoid having to make each runner a required status check). 

As a result, the `vrt` workflow lists the runners as a dependency and will correctly set the job status based on the runners. Unfortunately, the default would be for `vrt` to _not_ run if one of the runners fail so we list `always()` here to make sure this runs regardless of if a runner fails so that it can report the job status.

lukasoppermann:
Hey @joshblack, so is this a working version of it already? Or does it need more work?

In any case, feel free to leave it here as a draft until it can get merged. Thanks for the help 🙏

langermank (approved):

PR #640, opened 2023-05-03
Removing source files for never released v2 tokens

lukasoppermann:
## Summary

V2 tokens were never really released so they don't need to be deprecated. 

## Todo before merging
- [ ] @rezrah is using the tokens in primer/brand, so we can't delete them before primer/brand is updated
- [ ] Release v8 tokens
- [ ] Replace all v2 with v8 tokens (e.g. [PVC](https://github.com/search?q=repo%3Aprimer%2Fview_components+--borderWidth&type=code), [PCSS](https://github.com/search?q=repo%3Aprimer%2Fcss+--borderWidth&type=code), [dotcom](https://github.com/search?q=repo%3Aprimer%2Fcss+--borderWidth&type=code))


simurai (commented):
> V2 tokens were never really released so they don't need to be deprecated.

Yeah, we never "officially released" the v2 tokens. But we already started to use them in [PVC](https://github.com/search?q=repo%3Aprimer%2Fview_components+--borderWidth&type=code), [PCSS](https://github.com/search?q=repo%3Aprimer%2Fcss+--borderWidth&type=code) and even on [dotcom](https://github.com/search?q=repo%3Aprimer%2Fcss+--borderWidth&type=code).

It should be fine to remove the v2 tokens, since hopefully all cases have a fallback (e.g. hard coded `1px`). Although if feels more correct to:

1. Release v8 tokens
2. Replace all v2 with v8 tokens
3. Remove v2 tokens from Primitives.

/cc @langermank

langermank:
@simurai @lukasoppermann I already removed all references to v2 in dotcom/PCSS/PVC/PRC and point them to v8 (the secret pre-release) 😄 the only thing holding this back is Brand, but we haven't heard back from Reza yet if that's an issue.

rezrah:
Thanks for the ping, and sorry for the late response. 

I'm likely missing some background on this change, could I ask why these need to be removed right now? V2 tokens are still a core dependency for Primer Brand, so I'm a little worried that removing this will effectively cut the library off from future updates. I'd also speculate that migration effort from v2 to v8 is quite substantial, which we haven't factored in this quarter. 

If you need to go-ahead with this, no worries.. we can look to prioritise upgrading with urgency 👍 ... alternatively, you could leave them in for now and I can remove these files after the Primer Brand upgrade is complete.

lukasoppermann:
@rezrah I think we don't need to rip it out immediately.

How about this:
1. We don't compile tokens anymore and remove them (compiled tokens) from the npm package
2. We remove the build script from `package.json`
3. We move `build.js` to `scripts/buildv2.js` (@rezrah you would need to update to use this file than, but it is a tiny change)
4. We remove any workflows that check v2 tokens adn remove v2 builds, etc. from workflows

This gives you some time to update brand and once you have it updated we can remove the tokens & pipeline.

PR #654, opened 2023-05-14
Fallbacks for postCss

lukasoppermann:
## Summary

This adds the fallback json files for PostCss

## Output

All files are output as they are apart from color, which are combined to themes.

E.g. 
- `base/size/size.json`
- `base/typography/typography.json`
- `functional/size/border.json`

`base/size.json`
```js
{
  "--base-size-4": "0.25rem",
  "--base-size-8": "0.5rem",
  "--base-size-12": "0.75rem",
  "--base-size-16": "1rem",
  "--base-size-20": "1.25rem",
  "--base-size-24": "1.5rem",
  "--base-size-28": "1.75rem",
  "--base-size-32": "2rem",
  "--base-size-36": "2.25rem",
  "--base-size-40": "2.5rem",
  "--base-size-44": "2.75rem",
  "--base-size-48": "3rem",
  "--base-size-64": "4rem",
  "--base-size-80": "5rem",
  "--base-size-96": "6rem",
  "--base-size-112": "7rem",
  "--base-size-128": "8rem"
}
```

## Questions
- [ ] is this the output we need? @langermank 
- [ ] Do we really not want colors? Why @langermank? If so, I will remove this from the output.


langermank (approved):
This is AMAZING! THANK YOU!

I don't mind if you want to leave the color files in, but we won't use it for now. We only need color fallbacks to test the new color tokens against the existing tokens, meaning we use another CSS var in the fallback instead of a raw value. 

To use raw colors, we would need to configure the PostCSS plugin to respect themes. I don't think we need that right now :)

lukasoppermann:
@langermank in this case I will see if I can easily remove colors (should be fine). I am not a big fan of adding unused files. 👍

PR #661, opened 2023-06-14
prototype presentational collection

lukasoppermann:
## Summary

This PR adds presentational colors.

To make sure we don't have any impact on existing infrastructure (yet) I did not connect those scales to the base scales.
All colors are separate from base scales.

## List of notable changes:
- **added** new design token for presentational colors


langermank:
Closing as we've added these in a separate PR

PR #664, opened 2023-06-23
WIP: script to create a property-variable map for autocomplete & stylelint

siddharthkp:
(Do not merge, see demo in https://github.com/github/primer/issues/2008#issuecomment-1604286039)

This is a first draft, the implementation is based on running filter functions on generated variables (`tokens-next-private/**/*.css` 

Example:

```jsx
paddingBlock: ({type, name, file}) => {
    if (file.name === 'size.css') {
       if (type === 'base') return true
       else if (name.includes('padding') && !name.includes('paddingInline')) return true
    }
},
```

generated output: 

```json
{
  "paddingBlock": [
    { "name": "--stack-padding-normal", "value": "1rem" },
    { "name": "--stack-padding-spacious", "value": "1.5rem" },
    { "name": "--base-size-4", "value": "0.25rem" },
    { "name": "--base-size-8", "value": "0.5rem" }
  ]
}
```

[see full file](https://unpkg.com/browse/@primer/primitives@0.0.0-20230628151933//tokens-next-private/js/intellisense.js)

----

I'd like to explore adding property information to each token and treating intellisense as a [`platform`](https://github.com/primer/primitives/tree/main/src/platforms) as an alternative for this script

example:

```diff
 {
  "control": {
    "stack": {
      "padding": {
        "condensed": {
          "$value": "{base.size.8}",
          "$type": "dimension",
+         "$property": "padding"
        },
        "normal": {
          "$value": "{base.size.16}",
          "$type": "dimension",
+         "$property": "padding"
        },
        "spacious": {
          "$value": "{base.size.24}",
          "$type": "dimension",
+         "$property": "padding"
        }
      }
    }
  }
}
```


lukasoppermann:
Hey @siddharthkp to make the figma token plugin work, I am adding scopes to tokens: https://github.com/primer/primitives/pull/665/files#diff-a22f7c98b98666dfd7ade4b02294caee8eef00a58a4c80b0584f845ec3383d2eR10

I need specific scopes but I am sure we could work something out that works for both cases.

Any additional meta data should be added into the `$extension` property according to the w3c specs.

I currently nest my stuff into a `org.primer.figma` prop, but we can rename this to work for both our cases.

This is what I currently use:

```
$extension: {
        'org.primer.figma': {
          collection: 'mode',
          mode: 'dark',
          scopes: ['fgColor'],
        },
      },
```

The scopes I have are:

```js
  all: ['ALL_SCOPES'],
  radius: ['CORNER_RADIUS'],
  size: ['WIDTH_HEIGHT'],
  gap: ['GAP'],
  bgColor: ['FRAME_FILL', 'SHAPE_FILL'],
  fgColor: ['TEXT_FILL'],
  borderColor: ['STROKE'],
```
I will need to be able to keep up the mapping in a way so that the output stays the same, but we can add more granular scopes, e.g. `padding`, `margin`, `gap` could all map to `GAP` in figma but stay separate for your case.

siddharthkp:
Ooooh, that seems like the one!

> but we can add more granular scopes, e.g. padding, margin, gap could all map to GAP in figma but stay separate for your case.

That would be perfect!

siddharthkp:
> I currently nest my stuff into a org.primer.figma prop, but we can rename this to work for both our cases.

I'd suggest we start in another extension key, something like "org.primer.vscode", and then we can merge them once we have confidence in the output?

lukasoppermann:
> I'd suggest we start in another extension key, something like "org.primer.vscode", and then we can merge them once we have confidence in the output?

Great idea. 👍

PR #665, opened 2023-06-26
Figma tokens build process

lukasoppermann:
## Summary

This PR adds a build step to prepare primitives to be consumed by figma.

### Todo
- [x] create json file with all files listed, so that they can be fetched
- [x] add transformers to transform tokens in a way that figma can consume
- [x] add collection data to tokens so that they can be added to the correct collection and mode in figma
- [x] evaluate how to best prepare files (e.g. should all themes be in one large file?)


langermank (approved):
Looks good! Not going to read through the token src files but I think if there was a bug we would see a test failure 😄 🚀

PR #676, opened 2023-07-05
Disable unused imports

lukasoppermann:
## Summary

When releasing v8 we don't want user to use any platform other than `css` with `css variables`.
Not offering alternative platforms is the best way to avoid this.

## List of notable changes:

- **disabled** typeDefinitions
- **disabled** scss
- **disabled** js
- **disabled** json (not the json for docs)
- **disabled** typescript

- any reason why any of those platforms are needed?

### DIFF
--- a/scripts/buildTokens.ts
+++ b/scripts/buildTokens.ts
@@ -2,14 +2,14 @@ import type StyleDictionary from 'style-dictionary'
 import {PrimerStyleDictionary} from '~/src/PrimerStyleDictionary'
 import {copyFromDir} from '~/src/utilities'
 import {
-  typeDefinitions,
+  // typeDefinitions,
   deprecatedJson,
   css,
   docJson,
-  scss,
-  javascript,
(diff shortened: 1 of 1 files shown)


siddharthkp:
> any reason why any of those platforms are needed?

We may or may not want typescript depending on our implementation for primer/react. Happy to ship this PR now and bring typescript back if needed

lukasoppermann:
> > any reason why any of those platforms are needed?
> 
> We may or may not want typescript depending on our implementation for primer/react. Happy to ship this PR now and bring typescript back only if needed

Currently it is only disabled so bringing it back is easy. From our discussion yesterday I feel the best approach is to just use css as we want to move towards a future where we don't have lookups in react when styling.

langermank:
I think we should keep scss for now as we might want to try and introduce it to Primer CSS as another entry point towards using Primitives. 🙌

langermank (approved):

PR #682, opened 2023-07-07
Support `inverse` color tokens for Toast and Tooltip

langermank:
Toast design and color exploration. Drafting a concept of `inverse` tokens which handle dark backgrounds in light modes, and light backgrounds in dark mode. 

| Light mode | Dark mode |
| -- | -- |
| Light mode toast | Dark mode toast |

### New color tokens

Overlay color tokens are used by Dialogs, menus (ActionMenu, SelectPanel), Toasts, and Tooltips. Toast and Tooltip use `inverse` variants.

**Option 1: Overlay pattern tokens**

```
--overlay-inverse-bgColor
--overlay-inverse-fgColor
--overlay-inverse-fgColor-accent
--overlay-inverse-fgColor-success
--overlay-inverse-fgColor-attention
--overlay-inverse-fgColor-danger
```

**Option 2: Generic tokens for variant**

```
--overlay-inverse-bgColor
--overlay-inverse-fgColor
--fgColor-accent-onEmphasis
--fgColor-success-onEmphasis
--fgColor-attention-onEmphasis
--fgColor-danger-onEmphasis
```

Option 2 requires that we change the existing generic tokens like this:

`--fgColor-accent` -> `--fgColor-accent-default`

**Option 3: No inverse tokens, use data-attributes**

For this option, the components set the color mode data-attributes in the markup. Then we continue to use these color tokens (same we use for Tooltip today). So no pattern level tokens are needed (I think).

```
--fgColor-onEmphasis
--bgColor-emphasis
```


lukasoppermann:
Hey, interesting idea.

I feel the white alert is to high contrast in the dark theme. It seems like quite the rabbit whole with all the inverse colors you added.

What is the idea here? Do we need an inverse for every semantical color? Or just the 4 states and default?

Will this only be used in memex or will there be a lot more toasts coming up soon?
I am not 100% sure how I feel about the inverted style. I appreciate that it is more visible, but on mac OSX for example it also works with banners that are not inverted. 

If we add a lot more in the future I fear this may be too much high contrast elements. WDYT?

lukasoppermann:
So to get `--fgColor-accent-onEmphasis` and `--fgColor-accent` we need to change the tokens to this:

```js
fgColor: {
  accent: {
    "@": { 
      // this is the default value
    },
    "onEmphasis": {
       // here goes the stuff for the on emphasis value
     }
  }
}
```

I will adjust the transformers to remove the `@` from any name and this will result in the desired output.

Are we fine with this approach? If there are no concerns I will start working on it.

lukasoppermann:
Hey @langermank it passes the tests now.

It was a combination of both. An issue with the @-hack in references in the schema validation and you missed updating some references with the `@` and had one `$value: {}` which failed as well.

langermank:
Commenting to preserve the initial PR readme 😄 

We decided to move forward with extending primitive tokens to support an `inverse` scenario instead of using `overlay` specific colors (see token diff).

This PR also introduces some `overlay` size pattern tokens to support all overlay type components such as Toast, Tooltip, Overlay, Dialog, ActionMenu... etc etc. These are [values](https://primer.style/design/components/dialog#size) that are hard-coded today in CSS (I didn't just make them up).

lukasoppermann (approved):
The solution is great. I think having dedicated tokens is the cleanest way to do it and can be extended if needed. 👍 

Please make sure it all compiles well. I checked the files and it looks good, but I did not run style dictionry, so make sure the output is fine.

PR #726, opened 2023-09-07
ci(release): change variable and secret names for release

joshblack:
We ran into an error over at: https://github.com/primer/primitives/actions/runs/6106288510/job/16571101152 in the release workflow where the error indicated that the app id was not provided. This PR is adding in the shared variables and secrets in use in other repos that I _think_ are correct but wanted to double-check with you @camertron to see if there is another set we should be using 👀

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,11 +36,11 @@ jobs:
       - id: get-access-token
         uses: camertron/github-app-installation-auth-action@v1
         with:
-          app-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_ID }}
-          private-key: ${{ secrets.PRIMER_ISSUE_TRIAGE_APP_PRIVATE_KEY }}
-          client-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_CLIENT_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,11 +36,11 @@ jobs:
       - id: get-access-token
         uses: camertron/github-app-installation-auth-action@v1
         with:
-          app-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_ID }}
-          private-key: ${{ secrets.PRIMER_ISSUE_TRIAGE_APP_PRIVATE_KEY }}
-          client-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_CLIENT_ID }}
-          client-secret: ${{ secrets.PRIMER_ISSUE_TRIAGE_APP_CLIENT_SECRET }}
-          installation-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_INSTALLATION_ID }}
+          app-id: ${{ vars.PPRIMER_APP_CLIENT_ID_SHARED }}
```
jonrohan, line 39:
```suggestion
          app-id: ${{ vars.PRIMER_APP_CLIENT_ID_SHARED }}
```

camertron:
Thanks! This should already be fixed in main 👍

PR #728, opened 2023-09-12
update diffStat tokens

lukasoppermann:
## Summary

This PR removes `diffStat.addition.bgColor` as it can be replaced with `bgColor.success.emphasis`.
It also adds two borderColors for diffStat.

@langermank it would be even better to just remove the outline from red and green boxes, as they are not needed and barely visible. Do we own the component in Primer and can do this?


langermank:
Can you update the `color-fallback` to reflect this change? `"--diffStat-addition-bgColor": "var(--color-diffstat-addition-bg)",`

lukasoppermann:
Can be merged once https://github.com/github/github/pull/288487 is deployed

lukasoppermann:
@langermank fallback is now updated:

```diff
- "--diffStat-addition-bgColor": "var(--color-diffstat-addition-bg)",
+ "--diffStat-addition-bgColor": "var(--bgColor-success-emphasis)",
 ```

PR #730, opened 2023-09-12
Fix borderColor Success

lukasoppermann:
## Summary

Changes `borderColor/Success` to be the same as text which is better for outlined labels


langermank (approved):
Nice!

PR #734, opened 2023-09-19
Add icon color for invisible button

langermank:
- Adds `iconColor` tokens for the `invisible` button variant

| Before | After |
| -- | -- |
| icon button hover | icon button hover |


src/tokens/functional/color/light/patterns-light.json5
```diff
@@ -1093,6 +1093,44 @@
...
+            'org.primer.figma': {
+              collection: 'mode',
+              mode: 'light',
+              group: 'component',
+              scopes: ['fgColor'],
+            },
+          },
+        },
+        hover: {
+          $value: '{fgColor.muted}',
```
lukasoppermann, line 1110: In dark mode this is using `fgColor.default` shouldn't it be the same in both modes?

langermank: Spoke offline- it looks pretty harsh in light mode and isn't required to pass contrast, so I'm only changing the color in dark mode. I think it feels more expected to see a lighter effect in a dark theme on hover vs. darker in light mode.

PR #737, opened 2023-09-22
Fix npm release

rezrah:
## Summary

🚧 WIP

Fixing the library release to npm by using our Primer org workflow instead of a custom, local one.

## What approach did you choose and why?
- [Last two releases have failed](https://github.com/primer/primitives/actions/workflows/release.yml)
- I suspect they failed because the custom access token doesn't have the correct permission scopes for this repo, as [described here](https://github.com/changesets/changesets/issues/795#issuecomment-1631250128).
- Using the proposed token should work because it works for other Primer libraries

- We'll need to test in prod after merging by manually triggering the release. Added a command to do this via GitHub UI.

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - 'main'
       - 'next_major'
+  workflow_dispatch:
+
 jobs:
   release:
     name: Final
@@ -33,15 +35,6 @@ jobs:
       - name: Build experimental tokens
         run: npm run build:tokens
 
-      - id: get-access-token
-        uses: camertron/github-app-installation-auth-action@v1
-        with:
-          app-id: ${{ vars.PRIMER_APP_ID_SHARED }}
-          private-key: ${{ secrets.PRIMER_APP_PRIVATE_KEY_SHARED }}
-          client-id: ${{ vars.PRIMER_APP_CLIENT_ID_SHARED }}
-          client-secret: ${{ secrets.PRIMER_APP_CLIENT_SECRET_SHARED }}
-          installation-id: ${{ vars.PRIMER_APP_INSTALLATION_ID_SHARED }}
-
       - name: Create release pull request or publish to npm
         id: changesets
         uses: changesets/action@v1.4.1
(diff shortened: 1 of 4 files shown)


langermank (approved):
Thank you 🙌

camertron (changes requested):
Thanks for looking into this @rezrah, but I'm hesitant to merge these changes. I just finished a bunch of work to move away from `GPR_AUTH_TOKEN_SHARED` in favor of a GitHub App. The trouble with `GPR_AUTH_TOKEN_SHARED` is that it will expire over time, whereas the tokens issued by an App are sort-lived and issued on an as-needed basis. I believe the problem @langermank ran into yesterday is due to data missing from a GitHub API response, not an issue with the access token or the permissions attached to it. I'm currently looking into a fix.

rezrah:
> Thanks for looking into this @rezrah, but I'm hesitant to merge these changes. I just finished a bunch of work to move away from `GPR_AUTH_TOKEN_SHARED` in favor of a GitHub App. The trouble with `GPR_AUTH_TOKEN_SHARED` is that it will expire over time, whereas the tokens issued by an App are sort-lived and issued on an as-needed basis. I believe the problem @langermank ran into yesterday is due to data missing from a GitHub API response, not an issue with the access token or the permissions attached to it. I'm currently looking into a fix.

Absolutely fine 👍. Will close this and allow you to investigate, as i'll be unable to look into it further until Monday now. 

> ran into yesterday is due to data missing from a GitHub API response

Inclined to agree with this, as it could be related to the commit history thing raised in the changeset support issue. Have you tried [backing out this commit](https://github.com/primer/primitives/commit/0d982b1c7629afcdaa0a3168ffcfbfbae17aeeee) and seeing if that fixes anything?

camertron:
@rezrah I see the same failure for both our release.yml and release_canary.yml workflows. The canary one doesn't even get an auth token 🤔 I haven't tried backing out the commit, but that's a good idea. Thanks!

camertron:
Turns out this was caused by a git hash collision. I was able to fix things in the short term by reverting [the offending commit](https://github.com/primer/primitives/commit/51dbba881eea1cd1a2d0c601023c9d4b8357b64d), then reverting the revert. I've also submitted [a PR to the changesets project](https://github.com/changesets/changesets/pull/1236) to get this fixed long-term.

PR #744, opened 2023-10-03
Introduce tokens to support loading patterns

mperrotti:
## Summary

The primary goal of this PR was to add a design token to represent the amount of time we should wait before rendering a loading indicator or announcing a loading state to assistive technologies.

I also took some time to add other tokens I know we'll need to support the loading spinner and the "skeleton" loader.

### Added
- Duration to wait before showing a loading indicator
- Duration it takes the [Spinner](https://primer.style/design/components/spinner) indicator to complete a 360deg rotation
- Duration it takes the skeleton loader's "shimmer" to go from left to right
- Background color for skeleton loaders
- Dimensions for the different Spinner sizes
- Stroke width of the Spinner

- Does this taxonomy make sense?
- Do these feel like the right tokens to add?

## Steps to test:

TBD

- https://primer.style/design/components/spinner
- https://github.com/primer/design/pull/590

## Contributor checklist:

## Reviewer checklist:


src/tokens/functional/color/dark/patterns-dark.json5
```diff
@@ -1933,6 +1933,12 @@
       },
     },
   },
+  skeletonLoader: {
+    bgColor: {
+      $value: '{bgColor.muted}',
```
mperrotti, line 1938: Do we need to add an entry for dark mode if it's using the same token?

We use the same color between light and dark. We rely on `bgColor.muted` changing between dark and light modes.

src/tokens/functional/color/light/patterns-light.json5
```diff
@@ -1931,6 +1931,12 @@
       },
     },
   },
+  skeletonLoader: {
```
mperrotti, line 1934: I didn't add color tokens for the spinner because we just use `currentColor` with opacity. In most cases, this will mean the "track" of the spinner is `fgColor.default` @ 25% opacity, the bit that spins around the track is `fgColor.default` @ 100% opacity.

src/tokens/functional/color/dark/patterns-dark.json5
lukasoppermann: Yes, we need dark and light mode, just not any of the sub-themes.

src/tokens/functional/motion/patterns.json5
```diff
@@ -0,0 +1,26 @@
+/**
+ * Motion patterns
+ * @description Used to create animated components
+ */
+{
+  motion: {
+    spinner: {
+      rotationDuration: {
```
lukasoppermann, line 8: I am not sure, but WDYT think about a more general approach like `spinner.speed` and `skeletonLoader.speed`? 

CC: @mperrotti @langermank

lukasoppermann (approved):
One comment / idea about a more general approach for speed

src/tokens/functional/motion/patterns.json5
langermank: I feel like `duration` is different from `speed` in animation

langermank (approved):
Looks good! I would add a changeset for this with a patch, and then hopefully we'll resolve the build error so you can merge 😄

src/tokens/functional/motion/patterns.json5
lukasoppermann: Than maybe using just `duration` instead of `rotationDuration` and `shimmerDuration`?

lukasoppermann:
I updated this PR to main and now it has the duration validation support from #749 and the test is not failing anymore.

src/tokens/functional/motion/patterns.json5
mperrotti: I think `duration` could work for this case. When we start to add tokens for more complex animations, we might need to specify _what_ the duration is for.

For example: If I have an animation that scales in (`transform: scale()`) at a different rate than it fades in (`opacity`), we'll need to differentiate the two durations.

mperrotti:
I think we should add a transformer for duration tokens that strips the unit. So CSS would get `1000ms`, but JavaScript would just get `1000`.

What do you think?

lukasoppermann:
> I think we should add a transform for duration tokens that strips the unit. So CSS would get `1000ms`, but JavaScript would just get `1000`.
> 
> What do you think?

We currently don't build `js`. So, I would say if we at some point go back to building js as an output, we can add it. Does that make sense?

src/tokens/functional/motion/patterns.json5
lukasoppermann: Yes, true. But we could also think about nested naming? E.g. maybe this is part of the "scale"?

`duration.scale` and `duration.opacity`? 

Do you think we need to figure this out now, or should we go with `duration` and see if we ever run into issues?

lukasoppermann:
@mperrotti once you add the changeset it should be ready to merge. The `Show diff for v8 design token changes` is fine, as we are adding tokens, so the change is expected.

src/tokens/functional/motion/patterns.json5
langermank: So `motion-spinner-rotationDuration-default` could be `motion-spinner-duration-rotation`

with future tokens like..

`motion-spinner-duration-scale` or `motion-spinner-duration-opacity`

I like this.

mperrotti:
> We currently don't build js. 

We might be able to get around using `setTimeout` by relying on CSS delays. If not, we can go back and add it.

Also, we still style in JS in PRC (for now).

lukasoppermann:
> We might be able to get around using `setTimeout` by relying on CSS delays. If not, we can go back and add it.
> 
> Also, we still style in JS in PRC (for now).

This is changing with v8, right @langermank? That was why we removed js from the output?

langermank:
> Also, we still style in JS in PRC (for now).

@mperrotti @lukasoppermann this is still an open question and I'm writing up a discussion/ADR to pave a path forward (using JS vs. CSS vars)

PR #778, opened 2023-11-17
Update inactive button text color to meet contrast on inactive button bg color

mperrotti:
## Summary

Update inactive button text color in light and dark mode to ensure we're meeting the contrast ratio minimum against the inactive button's background color.

## List of notable changes:

- Updates inactive button colors in legacy and v8 token data

-

The former `fg.muted` value just barely fails the contrast ratio check when used on `gray.1`: https://github.com/primer/react/actions/runs/6893293489/job/18752536079?pr=3812

## Contributor checklist:

## Reviewer checklist:


mperrotti:
It looks like we might need to update `fg.default` as well. There isn't enough contrast between `muted` and `default` right now.

lukasoppermann:
> I think instead we should just use a custom color for the inactive button text tokens. What do you think?

yes, you are right. I missed the 3:1 contrast between default and muted, sorry. 

In this case we should go ahead and just create custom component token here.

langermank (approved):

PR #782, opened 2023-11-21
Handle `muted` foreground color combined with default control background in dark mode

langermank:
## Summary

In the default light theme, the `muted` or subtle canvas background color matches `button` (default control). In default dark, it's off by one step on the scale. We can either adjust that background color, or bring the `fgColor-muted` up a little bit.

### Option 1: adjust default control background color

| before | after |
| -- | -- |
| button before | button after |

### Option 2: adjust muted foreground color

| before | after |
| -- | -- |
|  button before | button after |

I'm leaning towards option 2 because it seems like the background color in dark mode was intentional (perhaps to add more contrast). If we had stronger border colors for controls, I wouldn't mind changing that background color.


lukasoppermann:
Hey @langermank I am in favor of changing the `fgColor-muted` as well. However we need to make sure that we preserve a `3:1` contrast between `fgColor-muted` and `fgColor-default`. Is this still `3:1` after the change?

maximedegreve:
I'm also in favour of option 2.

langermank:
@lukasoppermann not quite.. I'm not sure that would be possible 🤔 is this close enough? Let me know if not and I can keep manipulating things.

contrast ratio between default and muted fg colors in dark mode

lukasoppermann:
> @lukasoppermann not quite.. I'm not sure that would be possible 🤔 is this close enough? Let me know if not and I can keep manipulating things.
> 
> contrast ratio between default and muted fg colors in dark mode

Sadly in a11y it is either above or below the threshold. There is no _close enough_.

However I think since link underline preferences are shipped we don't require a 3:1 threshold anymore. Once confirmed we could remove them from the checks.

langermank:
For posterity, we decided to move forward with this change as we no longer require a 3:1 contrast ratio between `default` and `muted`. Previously, `muted` was the only way to indicate a link. Now we have the option to turn on underlines (which are on by default).

PR #785, opened 2023-11-22
Build v8 tokens to `/dist`

langermank:
Adding the v8 tokens to `dist` so we can start updating our documentation with the final import paths. Right now `dist` will be a combination of new and old stuff, and with our next major we'll remove the legacy build for just v8 tokens. 

I left the old build in place for backwards compatibility, which we can also remove with v8. 

Do I need to update any build processes to ensure the new tokens don't get deleted from `/dist` when the old tokens build?

### DIFF
--- a/.changeset/wise-hornets-sin.md
+++ b/.changeset/wise-hornets-sin.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Build v8 tokens to `/dist`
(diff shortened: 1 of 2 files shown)


jonrohan (approved):
> Do I need to update any build processes to ensure the new tokens don't get deleted from /dist when the old tokens build?

Not that I know of.

broccolinisoup (approved):
>Do I need to update any build processes to ensure the new tokens don't get deleted from /dist when the old tokens build?

Looks like both old and new tokens are combined under [dist](https://unpkg.com/browse/@primer/primitives@0.0.0-20231122225427/dist/) so I imagine it is fine

lukasoppermann (approved):
💪

PR #789, opened 2023-12-05
Addressing a few lingering bugs with v8

langermank:
- Fix foreground color for ToggleSwitch in dark high contrast
- Bring `borderColor-muted` closer to the existing color


lukasoppermann:
Just noticed the off state in Dark-HC on bgColor-muted. Do we mind or is this fine?

CleanShot 2023-12-05 at 09 18 08@2x

lukasoppermann (approved):
Great. Just added one comment.

PR #791, opened 2023-12-08
Add back default button shadow token

langermank:
Reverts part of https://github.com/primer/primitives/pull/733

screenshot of old PR

I was wrong.

### DIFF
--- a/.changeset/cuddly-gifts-sit.md
+++ b/.changeset/cuddly-gifts-sit.md
@@ -3,3 +3,4 @@
 ---
 
 Add back default button shadow token
+


joshblack (approved):

langermank:
@lukasoppermann sorry I was a bit lazy with this PR 😸 I noticed that the single React button in the global nav looks different from Rails, and that's because we removed this token. Seems tiny but I think we should be more careful about making shadow changes like that since it seems so intentional. Maybe we can revisit later

global nav buttons

langermank:
Closing this because I can't remember why I opened it 🤷

PR #793, opened 2023-12-21
Fixes for HC-dark

lukasoppermann:
## Summary

Currently all scales in Dark High Contrast have a bug where two colors are identical. This PR fixes that.

CleanShot 2023-12-21 at 10 43 19@2x


langermank (approved):
Thank you!

PR #805, opened 2024-02-16
Adding all color modes to figma export

lukasoppermann:
This PR adds transformations to be able to import all colors modes into Figma.

### DIFF
--- a/scripts/buildPlatforms/buildFigma.ts
+++ b/scripts/buildPlatforms/buildFigma.ts
@@ -48,7 +48,7 @@ export const buildFigma = (buildOptions: ConfigGeneratorOptions): void => {
       include,
       platforms: {
         figma: figma(`figma/themes/${filename}.json`, buildOptions.prefix, buildOptions.buildPath, {
-          mode: filename.replace('-', ' '),
+          mode: filename.replaceAll('-', ' '),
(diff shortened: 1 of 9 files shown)


scripts/buildPlatforms/buildFigma.ts
```diff
@@ -30,7 +38,7 @@ export const buildFigma = (buildOptions: ConfigGeneratorOptions): void => {
   }
   //
   for (const {filename, source, include} of themes) {
-    if (['light', 'dark'].includes(filename)) {
+    if (!['light-high-contrast', 'dark-high-contrast', 'dark-dimmed'].includes(filename)) {
```
langermank, line 41: What about color blind themes?

langermank (approved):

scripts/buildPlatforms/buildFigma.ts
lukasoppermann: This is just disabling them to test if they are problematic, they are. It is just a temp step.

lukasoppermann:
So it works great for all modes that have no specific base scale, so everything but `light-high-contrast`, `dark-high-contrast`, `dark-dimmed`.

So I need to find a fix for those three.

PR #810, opened 2024-02-23
Shadow color adjustments

langermank:
Adding a specific `overlay` color to be used within the `floating` shadows. This allows us to set a high contrast value for light/dark.


src/tokens/functional/shadow/light.json5
lukasoppermann, line 1: Do we want to have an effectively invisible border in light mode? I am fine either way, just asking.

PR #839, opened 2024-03-18
Add internalCss output

lukasoppermann:
## Summary

Creates a new output in `internalCss` which has color files in css that include the base color scales.


langermank (approved):
Thank you!

PR #851, opened 2024-04-01
rm most scales from output

lukasoppermann:
## Summary
This PR removes `color-scale-` colors from the output but keeps it in `internalCss`.

## Problem
We have some overwrites to change the scale colors, that are still present in the output.
However, if we want it to be present in the `internalCss` we need to add it somewhere.

I am wondering if we can remove just remove the `scale-color` colors from `internalCss` as well, since we also have the `base-color` tokens in `internalCss`. @langermank ?

I think we should be able to jsut rip out `color-scale` entirely if we don't need them in the output. They just duplicated the base scales without the need to include base scales.

Those are still available in `internalCss`:
CleanShot 2024-04-01 at 10 18 49@2x


lukasoppermann:
@langermank I updated storybook to use base colors and removed the color-scale colors. Is there any reason we can't do this?

PR #948, opened 2024-05-02
fix: Increase secondary text contrast

smockle:
## Summary

Fixes https://github.com/github/primer/issues/2216

This PR makes the “Dark” theme’s `fgColor.muted` slightly lighter. This improves secondary text contrast[^1] in hovered ActionMenu items:

| Element | Before | After |
|:---|:---|:---|
| Secondary Text | `#848d97`<br>75% neutral-3[^2] + 25% neutral-4 | `#8d96a0`<br>5% neutral-2 + 95% neutral-3 |
| Background (on hover) | `#292e36` | `#292e36` |
| Contrast | 4.06:1 | 4.56:1 |
| Screenshots | Before screenshot of ActionList, with a menu item hovered. Secondary text—“Due by March 31, 2022”—is slightly darker. | After screenshot of ActionList, with a menu item hovered. Secondary text—“Due by March 31, 2022”—is slightly lighter. |

- **Should we change text or background?** — I chose to lighten the secondary text color. Alternatively, to improve contrast, I could have darkened the hover background color. When I explored darkening the hover background color, I felt it became too hard to see which item was hovered.

- **What’s the right new color?** — The new color I chose—5% neutral-2 + 95% neutral-3—is arbitrary. There are many other colors that would have sufficient contrast. I picked `weight: 0.95` after reviewing the other `weight`s in `src/tokens/functional/color/dark/primitives-dark.json5`—each other `weight` in that file uses `0.05`-increments (`0.25`, `0.35`, etc.).

- **Will this break something else?** — I exclusively tested ActionMenu. I assume we use `fgColor.muted` in a lot of other places. How can we ensure this change won’t worsen contrast elsewhere?

## Steps to test:

1. Open this PR’s Storybook deployment, and navigate to the [ActionMenu > Examples > Groups and Descriptions story](https://primer.style/react/storybook/?path=/story/components-actionmenu-examples--groups-and-descriptions&globals=colorScheme:dark).
2. Open the ActionMenu (if it’s not already opened), then hover one of the items.
3. Inspect the contrast between secondary text (e.g. “Due by March 31, 2022”) and the hover background color, using a tool like [Pika](https://superhighfives.com/pika).

- This was discussed in the April 30th, 2024 Accessibility Design Office Hours: https://github.com/github/accessibility/issues/6005#issuecomment-2088729563 (Hubber access only)
- https://github.com/github/primer/issues/2216 (Hubber access only)
- [Understanding SC 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) (emphasis mine):
  > The minimum contrast Success Criterion (1.4.3) applies to text in the page, including placeholder text and **text that is shown when a pointer is hovering over an object** or when an object has keyboard focus. If any of these are used in a page, the text needs to provide sufficient contrast.

## Contributor checklist:

## Reviewer checklist:

[^1]: Specifically, contrast is increased above the 4.5:1 ratio required by [WCAG SC 1.4.3: Contrast (Minimum)](https://www.w3.org/TR/WCAG22/#contrast-minimum).
[^2]: These values are listed in [Primer Foundations > Color > Base Scales](https://primer.style/foundations/color/base-scales) (ensure “Dark” is selected). A quick way to test this in devtools is to use the CSS [`color-mix`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) function:
     ```css
     color: color-mix(in srgb, #b1bac4, #8b949e 95%)
     ```


lukasoppermann:
Hey, technically, since the contrast checks are passing this should not break anything. There is of course the possibility of tokens we don't have in our test matrix (including this very case with the hover).

We are working on moving away from using `mix`, however I guess for the time being it is a good change, since we already use `mix` anyway.

src/tokens/functional/color/dark/primitives-dark.json5
```diff
@@ -34,8 +34,8 @@
         },
       },
       mix: {
-        color: '{base.color.neutral.4}',
-        weight: 0.25,
+        color: '{base.color.neutral.3}',
+        weight: 0.95,
```
lukasoppermann, line 38: @smockle since this so close to 100%, did you try to just remove the mix all together?

smockle: I did, but it’s not quite light enough: `#8b949e` (`base-color-neutral-3`) on `#292e36` is 4.45:1.

smockle: If we _also_ adjust [`control.transparent.bgColor.hover`’s `alpha`](https://github.com/primer/primitives/blob/44c86c0d327cef5a7180baec9be614222ac8fdc7/src/tokens/functional/color/dark/patterns-dark.json5#L298) from 12% down to 11%, then `base.color.neutral.3` (without mixing) is sufficient.

lukasoppermann:
> > …since the contrast checks are passing this should not break anything. There is of course the possibility of tokens we don't have in our test matrix (including this very case with the hover).
> 
> Should I add a test for this? If so, would that just require adding a line to
> 
> https://github.com/primer/primitives/blob/44c86c0d327cef5a7180baec9be614222ac8fdc7/scripts/color-contrast.config.ts#L12
> 
> , like:
> ```js
> [4.5, 'fgColor-muted', 'control-transparent-bgColor-hover'],
> ```

Hey, no thank you, I will do this in a new PR and add more missing cases.

lukasoppermann (approved):

lukasoppermann:
@langermank now thinking about this, shouldn't the ActionMenu use `--control-fgColor-rest`? However, there is no equivalent to `--fgColor-muted`. I am just wondering when `control-rest` vs `fgColor-default` should be used.

owenniblock (approved):
Looks good to me 👍

PR #956, opened 2024-05-16
Removed space tokens

lukasoppermann:
## Summary

> [!CAUTION]
> This can only be considered a minor change if we remove any use of `--space` tokens before we merge and release this.

This PR removes the `--space` tokens. They are deprecated and users should use `--base-size` tokens instead.

However, they are still in use:
- [GitHub](https://github.com/search?q=org%3Agithub+var%28--space-+repo%3Agithub%2Fgithub&type=code)
- [Primer](https://github.com/search?q=org%3Aprimer+var%28--space&type=code) (only 2 cases)
  - https://github.com/primer/react/pull/4598

## Tasks
- [ ] ~update styleLint plugin: https://github.com/primer/stylelint-config/blob/86cf24f5c1f3f51b5085a5808a5406bc04e47b68/__tests__/__fixtures__/good/example.module.css#L131 also to suggest base size for 2px and 6px~
- [x] fix uses for [GitHub](https://github.com/search?q=org%3Agithub+var%28--space-&type=code)


lukasoppermann:
@langermank I removed the use, do you think we can merge this with just a minor version bump?

.changeset/orange-jokes-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/primitives': minor
```
lukasoppermann, line 2:
```suggestion
'@primer/primitives': major
```

langermank (approved):

PR #968, opened 2024-06-14
Replaces `--data-[colorName]-color` with `--data-[colorName]-color-emphasis`

lukasoppermann:
# Important
- replace any usage of `var(--data-[colorName]-color)` with `var(--data-[colorName]-color-emphasis, var(--data-[colorName]-color))` before merging.

→ [Github Code Search for variables](https://github.com/search?q=org%3Agithub+%2Fvar%5C%28--data-%5Ba-z%5D%2B-color%2F&type=code)

PR that updated `github/github` to use new tokens with old ones as fallback: https://github.com/github/github/pull/330119

## Summary

Figma uses every level of nesting in the JSON to create a folder structure.

This creates:

```
data/
  blue/
    color
    color/
      muted
  auburn/
    color
    color/
      muted
  orange/
    color
    color/
      muted
```

To avoid this, we need to rename in the json.


jonrohan (approved):

langermank (approved):
Dotcom has been updated, but still merging this into `major`

PR #976, opened 2024-06-20
New `light` and `dark` neutral scales

langermank:
Introducing... new `light` and `dark` neutral base scales! 🎉

neutral scales

This isn't technically a breaking change and there should be very little visual differences right now, but we'll be releasing this in Primitives v9 since its a new way of approaching neutral base scales.

### Inverted scales

Going forward, our base scales will follow a reverse pattern where `0` is the lightest value in `light` and the darkest value in `dark`. This makes it easier to maintain and use for functional tokens where in many cases, both light and dark themes will be identical in base value.

### Two scales to rule them all

Rather than maintaining separate high contrast scales, we now have only two scales that handle low, medium and high contrast. We adjust contrast by moving up and down the scale, like darker background shades or darker text shades for light high contrast. 

### Contrast ratios and application

Each step of the scale serves a purpose. The first 5 shades are primarily background shades, 6 and 7 are borders, and 8 to 11 are for text.


lukasoppermann:
Can we fix this in here or would it better be a separate fix?

**Hover state of button**
image

I think the fix would be a using `black` with a very low alpha, like `0.1` or even lower.

PR #1000, opened 2024-07-09
Update e2e test

lukasoppermann:
This is supposed to automatically use the latest installed version of playwright/test


joshblack (approved):
Looks like it works 🎉

PR #1009, opened 2024-07-10
Create adr-002-fgColor-onEmphasis.md

lukasoppermann:
See [ADR](https://github.com/primer/primitives/pull/1009/files?short_path=6ea6e72#diff-6ea6e720235f70970b500ad313de19e24582b48588db491ff8b2a51bc7df970a)


langermank:
I'd like to add some images and context to help us further discuss this decision. 

Currently, in dark high contrast theme we invert `fgColor-onEmphasis` from white to black. This is the only theme we do this in. The decision to do this doesn't appear to be recorded, so we don't know for sure the reasoning. We're assuming this was done so that background colors could be more saturated and appear higher contrast against the default/muted backgrounds.

The inverted color is a bit challenging from a maintenance perspective, and increases the overall complexity of our color system. However, if it provides a better or more comfortable reading experience I'm personally fine with keeping it.

Below is a snippet of the existing experience. I know its difficult to evaluate out of context but just to give us a rough idea. On the left is default dark theme with both `bgColor-default` and `bgColor-muted`.

existing dark compared to dark high contrast themes

This is a proposed path forward where we use white text, but maybe introduce more bright border colors to add more contrast. 

proposed theme comparing dark and dark high contrast

In the existing experience, the contrast ratios are as follows for the primary button:

| colors | ratio |
| -- | -- |
| `bgColor-muted` vs primary button bg | 6.3 |
| primary button bg vs primary button text | 7.15 |
| primary button border vs `bgColor-muted` | 10.23 |

In the proposed solution, the contrast ratios are as follows:

| colors | ratio |
| -- | -- |
| `bgColor-muted` vs primary button bg | 2.3 |
| primary button bg vs primary button text | 7.6 |
| primary button border vs `bgColor-muted` | 10.23 |

Without the border color, the background definitely loses some contrast against `bgColor-muted`. However, with white text it seems like we may actually increase the contrast.

Would love to hear some thoughts and opinions!

langermank (approved):
I made some edits, but I feel confident moving forward with this decision since we've been internally testing it with no complaints.

PR #1019, opened 2024-07-31
Add stories for all transparent colors

langermank:
Adds a set of stories displaying the breakdown of functional colors that have an `alpha` set.

table


siddharthkp (approved):
👍✨🎨

PR #1024, opened 2024-08-08
Extend neutral scale to include white/black (0-13)

langermank:
Changes the new neutral scales from 0-11 to 0-13, including white and black in the scale for easier functional color mapping.

neutral-scales


lukasoppermann (approved):
Love it. 👍 

Great idea to ref neutral/0 to white / black.

PR #1032, opened 2024-08-28
convert exported build config to js (used by brand)

lukasoppermann:
## List of notable changes:

-

-

## Steps to test:

<!--

2. Go to # page

1.
1.
1.

-

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,7 @@
     "node": ">=18.18.0 <18.19.0"
   },
   "files": [
-    "dist",
-    "src/**/*",
-    "!src/**/*.test.ts"
+    "dist"
   ],
   "repository": "https://github.com/primer/primitives",
   "keywords": [
@@ -26,11 +24,12 @@
   },
   "homepage": "https://github.com/primer/primitives#readme",
   "scripts": {
-    "build": "npm run build:v8 && npm run build:fallbacks && npm run build:figma",
+    "build": "npm run build:v8 && npm run build:fallbacks && npm run build:figma && npm run build:config",
     "build:next": "echo '\\033[31m 🛑 Use npm run build instead of npm run build:next\\033[0m'",
     "build:v8": "tsx ./scripts/buildTokens.ts",
     "build:fallbacks": "tsx ./scripts/buildFallbacks.ts",
     "build:figma": "tsx scripts/buildFigma.ts",
+    "build:config": "tsc -p buildConfig.tsconfig.jsonc",
     "tokenJson:check": "tsx scripts/diffThemes.ts && tsx scripts/diffTokenProps.ts",
     "contrast:check": "tsx scripts/color-contrast.ts",
     "format": "prettier --write  '**/*.{js,jsx,ts,tsx,md,mdx,css}'",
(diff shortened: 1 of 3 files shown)


buildConfig.tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+  "compilerOptions": {
+    "jsx": "react",
+    "target": "ES2016",
+    "module": "commonjs",
+    "lib": ["ESNext", "dom"],
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "strict": true,
+    "noImplicitAny": true,
+    "skipLibCheck": true,
```
lukasoppermann, line 11: @rezrah I had to add `"skipLibCheck": true,` to avoid an error:

> node_modules/vite/dist/node/index.d.ts:5:41 - error TS2307: Cannot find module 'rollup/parseAst' or its corresponding type declarations.
  There are types at '/Users/lukasoppermann/GitHub/primitives/node_modules/rollup/dist/parseAst.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
> 
> 5 export { parseAst, parseAstAsync } from 'rollup/parseAst';
>                                           ~~~~~~~~~~~~~~~~~
> Found 1 error in node_modules/vite/dist/node/index.d.ts:5

However, I read that this is more like a "hack". What do you think about this?

rezrah: 🤔 yeah, disabling that is quite problematic. To unblock you, maybe disable for now, but add a comment above that line along with a TODO to make sure we follow up and fix it. Changing to `bundler` is a common way to fix it, but i'm not sure what rollup is used for unless I dig into the relevant code.

rezrah: Comment could be something like this:

```suggestion
    // FIXME: Disabled in https://github.com/primer/primitives/pull/1032
    "skipLibCheck": true,
```

package.json
```diff
@@ -7,7 +7,7 @@
   },
   "files": [
     "dist",
-    "src/**/*",
+    "build",
```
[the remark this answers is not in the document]
rezrah: 👍 the industry defacto approach is just to add compiled code to a dist or lib folders. Perhaps it's best to add there under `helpers` or `build` folder? E.g. `dist/build` or `dist/configs`?

In terms of breaking changes, it's entirely up to you how you communicate breaking changes and use SemVer. As Primitives is OSS, technically _any_ removal of files in a new version is a breaking change, and should be communicated as such. A 'contract' was made with users the moment you made it publicly available. SemVer is one way to comm that, but there's other ways too. If you remove, or rename it in future but don't want to create a `major` update because it's more experimental, you could manually call it out in the Changelog like this: https://github.com/primer/brand/releases/tag/%40primer%2Freact-brand%400.35.0. This might be sufficient.

rezrah (approved):

buildConfig.tsconfig.json
lukasoppermann: Changing to `bundler` sadly does not solve it. 

I like the comment idea.

PR #1036, opened 2024-08-29
Adding pascalCase transform to be used by brand

lukasoppermann:
This PR adds a pascalCase transformer

### DIFF
--- a/.changeset/dry-walls-marry.md
+++ b/.changeset/dry-walls-marry.md
@@ -1,0 +1,5 @@
+---
+'@primer/primitives': minor
+---
+
+Adding a pascalCase trasnformer
--- a/src/utilities/toPascalCase.ts
+++ b/src/utilities/toPascalCase.ts
@@ -1,22 +1,13 @@
+import {filterStringArray} from './filterStringArray'
 import {upperCaseFirstCharacter} from './upperCaseFirstCharacter'
 
 export const toPascalCase = (string: string | string[]) => {
   if (!Array.isArray(string)) {
     string = [string]
   }
-  // match unsupported characters
-  const regex = /[^a-zA-Z0-9]+/g
-  // replace any non-letter and non-number character and split into word array
-  const stringArray = string
-    .filter(part => part !== '@')
-    .filter(Boolean)
-    .join(' ')
-    .replace(regex, ' ')
-    .split(' ')
+
   return (
-    stringArray
-      // remove undefined if exists
-      .filter((part: unknown): part is string => typeof part === 'string')
+    filterStringArray(string)
       // ucFirst all but first part
       .map((part: string) => {
         return upperCaseFirstCharacter(part)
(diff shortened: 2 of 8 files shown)


.changeset/dry-walls-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/primitives': minor
```
rezrah, line 2: Just curious how you determine the version. Why minor vs patch?

src/utilities/toPascalCase.ts
```diff
@@ -0,0 +1,26 @@
...
+export const toPascalCase = (string: string | string[]) => {
+  if (!Array.isArray(string)) {
+    string = [string]
+  }
+  // match unsupported characters
+  const regex = /[^a-zA-Z0-9]+/g
+  // replace any non-letter and non-number character and split into word array
+  const stringArray = string
+    .filter(part => part !== '@')
+    .filter(Boolean)
```
rezrah, line 12: Nit, but you could combine these

src/utilities/toPascalCase.ts
```diff
@@ -0,0 +1,26 @@
...
+    .split(' ')
+  return (
+    stringArray
+      // remove undefined if exists
+      .filter((part: unknown): part is string => typeof part === 'string')
+      // ucFirst all but first part
+      .map((part: string) => {
+        return upperCaseFirstCharacter(part)
+      })
+      .join('')
```
rezrah, line 24: Another nit (not blocking) but you could do all array manipulation inside the function body and just return the final output in the return to make this a bit clearer. Right now it seems like there's the same things happening in both the return and function body.

rezrah (approved):
Left some minor, non-blocking feedback.

Thanks for doing this @lukasoppermann 🙏

.changeset/dry-walls-marry.md
lukasoppermann: @rezrah I figure we are adding a feature here, so minor is more appropriate. Let me know if you have some good guidelines as to when to choose which version.

src/utilities/toPascalCase.ts
lukasoppermann: Yes, but I feel it is easier to read this way and performance is not really a bottle-neck.

You mean like this, right?

```js
 .filter(part => part !== '@' && Boolean(part))
````

src/utilities/toPascalCase.ts
lukasoppermann: I thought about this as well. Will look into it. 👍

.changeset/dry-walls-marry.md
rezrah: Yeah that makes sense. Thanks.

src/utilities/toPascalCase.ts
lukasoppermann: Actually ended up refactoring this a bit

PR #1040, opened 2024-09-02
Adding tests that verify that the entire build setup is working as expected

lukasoppermann:
## List of notable changes:

-

-

## Steps to test:

<!--

2. Go to # page

1.
1.
1.

-

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 docs/public/
 color-contrast-check.json
 blob-report
+integration/build
--- a/integration/build.test.ts
+++ b/integration/build.test.ts
@@ -1,0 +1,62 @@
+import {PrimerStyleDictionary} from '../src/PrimerStyleDictionary'
+import fs from 'fs'
+
+describe('PrimerStyleDictionary', () => {
+  const basePath = `./integration`
+  const extendedSD = PrimerStyleDictionary.extend({
+    source: [`${basePath}/tokens/**/*.json5`],
+    platforms: {
+      css: {
+        prefix: 'PREFIX',
+        transformGroup: 'css',
+        buildPath: `${basePath}/build/css/`,
+        files: [
+          {
+            options: {
+              showFileHeader: false,
+            },
+            destination: 'variables.css',
+            format: 'css/variables',
+          },
+        ],
+      },
+      advancedCss: {
+        prefix: 'PREFIX',
+        transformGroup: 'css',
+        buildPath: `${basePath}/build/css/`,
+        files: [
+          {
+            options: {
+              showFileHeader: false,
+            },
+            destination: 'advanced.css',
+            format: 'css/advanced',
+          },
+        ],
+      },
+    },
+  })
+
+  extendedSD.cleanAllPlatforms()
+  extendedSD.buildAllPlatforms()
+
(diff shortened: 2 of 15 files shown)


integration/build.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+  })
+
+  extendedSD.cleanAllPlatforms()
+  extendedSD.buildAllPlatforms()
+
+  it('it should transform with css/variables format', () => {
+    const output = fs.readFileSync(`${basePath}/build/css/variables.css`, 'utf8')
+    const expectedOutput = `:root {
+  --prefix-base-color-blue-500: #2C29FF; /* The primary color for interactive elements. */
```
rezrah, line 46: qq... should this test instead assert that the prefix isn't applied to base tokens and only functional ones?

integration/build.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  --prefix-fg-color-link: #2C29FF;
```
rezrah, line 47: Similar question for this one.. the input is `fgColor` but output is `fg-color`. Is that normals? I'd expect it to also be `fgColor` in output?

integration/build/css/advanced.css
rezrah, line 1: Should this file be in VC? It's a test artifact from how I'm reading it.

integrationbuild/css/variables.css
rezrah, line 1: Think this file is in the wrong folder @lukasoppermann ? `integrationbuild`

.gitignore
```diff
@@ -7,3 +7,4 @@ coverage/
 docs/public/
 color-contrast-check.json
 blob-report
+e2e/build
```
rezrah, line 10: Should this be..
```suggestion
integration/build
```

integration/build.test.ts
lukasoppermann: Why? This is not the expected behavior for this config.

integration/build.test.ts
lukasoppermann: This is currently just testing the default naming transforms. Here this is expected behavior. I will add our custom name transformers and then this would be the expected behavior.

integration/build/css/advanced.css
lukasoppermann: No it should, good catch. Issue is the gitignore change I missed and you caught. 🙏

integration/build.test.ts
rezrah: ah okay.. so there's another transformer that does that, i think that makes sense then 👍

integration/build.test.ts
rezrah: Based on https://github.com/primer/primitives/pull/1040/files#r1740667105 I think I got my answer. The opinionated rules are applied in a different stage in the build process, so this isn't representative of an end-to-end test.

integrationbuild/css/variables.css
rezrah: This file still seems to be here, is this expected or an oversight?

integration/build.test.ts
lukasoppermann: Yes, I will add tests for this as well. Those are the baselines. :D

PR #1046, opened 2024-09-05
Extend line-height value

langermank:
Because LINUX I guess


lukasoppermann:
@langermank is this something we should merge?

PR #1053, opened 2024-09-12
Update `skeleton` pattern to work on muted bg

langermank:
Updating the value so that its visible on muted and `overlay` backgrounds


camertron (approved):
Awesome, thank you!

PR #1100, opened 2024-12-05
Allow to keep some tokens in output as references

lukasoppermann:
## Summary

This PR makes changes to the way references are used in the css platform. Now, all referenced tokens that are in the output will be used as references, while base tokens that are not in the output will be converted to hex values.

## Impact

This work enables a lot of use cases related to customization including. 
**In short, this update allows us to change design tokens (e.g. colors, text sizes, contrast, etc.) in the browser (e.g. user settings) that will propagate through all of github.**

For example:

- A user could define the contrast they want for border colors in the settings
- A user could choose a specific color for e.g. `success` states

### DIFF
--- a/.changeset/khaki-kiwis-explain.md
+++ b/.changeset/khaki-kiwis-explain.md
@@ -1,0 +1,5 @@
+---
+'@primer/primitives': minor
+---
+
+Changing the way we use references in css output, to keep all references to tokens that are present in the final output
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@primer/primitives",
-  "version": "10.1.0",
+  "version": "10.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@primer/primitives",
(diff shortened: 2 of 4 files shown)


lukasoppermann:
Current bug:

Style dictionaries [`outputReferencesTransformed`](https://github.com/amzn/style-dictionary/blob/f694f672e2d033dc79aea88b05591204c65321a6/lib/utils/references/outputReferencesTransformed.js#L11) function does not work with object tokens (composite tokens) as far as I can see.

The output for shadows now has a color value that is: `{color.token.reference} [2 digits of hex8 for alpha value]` since this this not a valid color, the shadow is broken:

```css
    --shadow-floating-xlarge: 0px 0px 0px 1px #d1d9e0, 
       0px 56px 112px 0px var(--base-color-neutral-12) 52; /* <- this 52 is the problem */
```

lukasoppermann:
I got feedback from the Style Dictionary team, this has been fixed by this PR: https://github.com/amzn/style-dictionary/pull/1401 and will be release in version `4.3.0`

PR #1104, opened 2024-12-05
[ADR] Add prefix to design tokens

lukasoppermann:
## List of notable changes:

-

-

## Steps to test:

<!--

2. Go to # page

1.
1.
1.

-

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/contributor-docs/adrs/adr-005-token-prefix.md
+++ b/contributor-docs/adrs/adr-005-token-prefix.md
@@ -1,0 +1,30 @@
+# Prefix for primitives
+
+## Status
+
+| Stage    | Status      |
+| -------- | ----------- |
+| Approved | 🚧 |
+| Adopted  | <!-- 🚧 --> |
+
+## Context
+
+Currently primitives have no prefix, which means once converted to css they can be used like `--fgColor-default`. This is very convenient but could lead to naming collisions or confusion if folks use their own css variables e.g. is `--button-fgColor-default` a token coming from primitive or a local variable?
+
+To solve this we could add a prefix like `primer` or `product` to our tokens, resulting in css variables like `--primer-fgColor-default`.
+
+## Decision
+
+The following decisions need ot be made:
+
+1. Do we think that this is a legit problem that is serious enough to invest somewhat significant resources into?
+2. If yes, what is an appropriate prefix?
+3. If no, do we need some kind of guidance in primer.style, or are we good without it?
+
+### Impact
+
+- We will need to a) keep both prefixed and non-prefixed tokens for quite a while and b) work on removing all non-prefixed tokens from dotCom doing PRs and using stylelint.
+
+### Alternatives
+
+Not using a prefix and keeping it as is.


contributor-docs/adrs/adr-005-token-prefix.md
```diff
@@ -0,0 +1,30 @@
...
+
+Currently primitives have no prefix, which means once converted to css they can be used like `--fgColor-default`. This is very convenient but could lead to naming collisions or confusion if folks use their own css variables e.g. is `--button-fgColor-default` a token coming from primitive or a local variable?
+
+To solve this we could add a prefix like `primer` or `product` to our tokens, resulting in css variables like `--primer-fgColor-default`.
+
+## Decision
+
+The following decisions need ot be made:
+
+1. Do we think that this is a legit problem that is serious enough to invest somewhat significant resources into?
```
mperrotti, line 20: No

contributor-docs/adrs/adr-005-token-prefix.md
```diff
@@ -0,0 +1,30 @@
...
+2. If yes, what is an appropriate prefix?
+3. If no, do we need some kind of guidance in primer.style, or are we good without it?
```
mperrotti, line 22: What would the guidance be? I don't think CSS variable collisions are specific enough to Primer to need documentation.

contributor-docs/adrs/adr-005-token-prefix.md
langermank: We discussed prefix _at length_ when we first introduced our new set of tokens. I'm sure I could dig it up 😄 

I don't think we need a prefix for product tokens, especially because we _do_ use a prefix for `brand` which is where collisions are actually very likely to happen.

If we're worried about people defining _new_ tokens with the same names as our tokens, we can lint for that and not allow it using stylelint or eslint.

We simply don't have the resources right now to invest in a mass rename like we've done earlier this year, nor do I think it's a good idea.

lukasoppermann: >If we're worried about people defining new tokens with the same names as our tokens, we can lint for that and not allow it using stylelint or eslint.

I think this is the main problem.

I do agree with the resources. However, the topic came up again and I want to have our decision documented, hence this ADR.

@langermank could you quickly outline how a linting against this would work, so that we can add it as an alternative to the ADR?

contributor-docs/adrs/adr-005-token-prefix.md
lukasoppermann: > What would the guidance be? I don't think CSS variable collisions are specific enough to Primer to need documentation.

Idk. Maybe suggesting using a custom prefix if you create your own tokens? Or outlining all the token types we have so that people can avoid conflicts.

But as I mentioned above, I am not necessarily a _"fan"_ of adding a prefix (mainly because of the significant resource demand). And I also don't really have a point for 3. yet. Just wondering if there is any.

contributor-docs/adrs/adr-005-token-prefix.md
langermank: Can you share where this is coming up? I haven't personally seen it discussed or encountered open issues so I'm curious 👀

lukasoppermann: This came up in a conversation I had Reza. 

I think if we have no open issues and since there is a huge amount of work required it would be fair to keep it as is.
However, we should document this decision in an ADR. Thats exactly what they are for, right?

rezrah: 👋 Sorry I'm late here, for context I raised this as @lukasoppermann shared an example [of the upcoming component-level variable format](https://github.com/primer/primitives/blob/dae7eab3c339aea29e167bcb0cb60069abf05287/src/tokens/component/header.json5) which are **already** prefixed for Brand, **but won't be for product.** 

Extrapolating from that example:

Primer Brand:  `--brand-headerSearch-bgColor`
Primer Product: `--headerSearch-bgColor`
Feature teams: `--headerSearch-customThing-bgColor`

My anecdotal thoughts on the above were:

- It's very easy to distinguish (in source code and debugging tools) between Primer Brand-owned component tokens from those custom-implemented by feature teams based due to the prefix
- It's not so easy to do the same with Primer product-specific tokens _for component-specific variables_
- Component tokens are (IMO) different to low-level primitives like `fgColor` because they contrastingly lead to an exponential increase in CSS variables creation. With the move to CSS modules, feature teams will inevitably create many new variables, which could exacerbate the opportunities for collision and confusion.

This is admittedly speculative, and largely based on my recent experience working on the Copilot project, where the team created many component-level CSS variables. 


It's one of those can-of-worms that easier to 'contain' now rather than later, as **you could prefix component-scoped tokens in Primer.**

> If we're worried about people defining new tokens with the same names as our tokens, we can lint for that and not allow it using stylelint or eslint.

👍 That's fair, though I think the ownership ambiguity and indirection would still be there. It also takes agency away from feature teams to author their own arbitrary tokens, when its relatively easy for Primer to prefix.

> We discussed prefix at length when we first introduced our new set of tokens. I'm sure I could dig it up 😄

Nice. Sorry for drudging up an old conversation. This arose from a conversation I had with @lukasoppermann, but will defer to y'all on if/how you wanna address it. We're also not blocked by this on the Brand side as we have the prefix in place today ✌️

lukasoppermann:
@mperrotti @langermank I updated the ADR to reflect this decision. Do you feel this makes sense now?

contributor-docs/adrs/adr-005-token-prefix.md
langermank: Thanks for clarifying @rezrah! It sounds like you're referring specifically to `component` level tokens which I think is interesting to consider. When we shipped tokens, we included the option for a `prefix` specifically for Brand, but we also had a `primer` prefix originally in the first set. We later decided to remove it as it felt redundant. I can see how this is shifting with the move towards CSS Modules.

The naming convention allows for `prefix` and also `namespace` which we only use for `base` tokens right now. If we were to explore a "primer" namespace for component/pattern tokens, that could work for both Product and Brand UI:

```
--pc-control-bgColor-rest
--brand-pc-control-bgColor-rest
```

Using `pc` as Primer Component but idk 😄 

So this model excludes functional tokens because this is not an area teams should be adding to or customizing. `--fgColor-default` is what it is and it would be redundant for teams to be introducing new tokens for established patterns. I do understand why defining tokens for components is useful, though. 

I believe in the original naming convention conversations, we felt that the component/pattern name itself was unique enough to not need a prefix. But I'm not sure that's true anymore.

lukasoppermann: @langermank this is a very interesting position.

we could also do:

`--primer-component-control-bgColor-rest`
`--brand-component-control-bgColor-rest`
or
`--brand-control-bgColor-rest`

While it is longer, I feel it is easier to understand than `pc`. 

@langermank do you think we should add this as a decision to the ADR and merge it?

If so, next steps would be to add an issue to track this work, and add this issue to the stretch goals.

PR #1193, opened 2025-03-11
Adding data gradient

lukasoppermann:
## Summary

This PR adds gradients for data visualization using the muted color.

I am adding special colors `gradientStop-1` and `gradientStop-2` as well as an assembled gradient. This is important because both in svg gradients and in Figma we need the individual colors rather than a combined gradient like in css.


src/tokens/functional/color/data-vis.json5
```diff
@@ -26,11 +26,47 @@
               scopes: ['bgColor'],
             },
             'org.primer.overrides': {
-              dark: '{base.display.color.blue.0}'
-            }
+              dark: '{base.display.color.blue.0}',
+            },
+          },
+        },
+        'gradient-stop-1': {
```
langermank, line 33: Is there a reason this is the full string instead of nested values for the token name?

lukasoppermann: Good point, I adjusted it to `gradientStop` and nested `1` and `2` inside.

PR #1198, opened 2025-03-20
Update workflow to update snapshots to use runners for faster builds

joshblack:
Porting over the logic we use for Primer Docs in: https://github.com/github/primer-docs/blob/main/.github/workflows/vrt.yml to primitives to help speed up how long it takes to update snapshots.


langermank (approved):
Lets see if this solves it! 🤞

PR #1203, opened 2025-03-27
Update playwright.config.ts to set retries to 0

joshblack:
Update our playwright setup so that the number of retries is 0. This will reduce the total time it takes a failing job to finish by up to 3x (one attempt plus two retries). The cost of this is that a job may fail if there is some flakiness that would be fixed by a retry.


langermank (approved):
Thank you!

PR #1220, opened 2025-05-20
Use `bgColor-default` for `overlay` background in dark themes

langermank:
Use `bgColor-default` for dark overlays. I'm really not sure how I feel about this solution as we rely on a slightly lighter shade to show dimension for dark mode where shadows aren't very useful. But we're lacking the ability to layer inside overlays by using `muted` as the background.

Closes https://github.com/github/primer/issues/4175

Before

Overlay demo before

After

Overlay demo after

### DIFF
--- a/.changeset/ten-adults-matter.md
+++ b/.changeset/ten-adults-matter.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": minor
+---
+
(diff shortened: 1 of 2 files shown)


lukasoppermann:
Can we make the BG darker so that we still see layers?

lukasoppermann (approved):

PR #1237, opened 2025-07-09
Add `label` border-color tokens

langermank:
Add border-color tokens to support high contrast themes.

### DIFF
--- a/src/tokens/component/label.json5
+++ b/src/tokens/component/label.json5
@@ -101,6 +101,46 @@
           },
         },
       },
+      borderColor: {
+        $value: 'transparent',
+        $type: 'color',
+        $extensions: {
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Add a storybook demo file similar to DisplayColorTesting.stories.tsx where you preview all label colors using a label component.

lukasoppermann:
It would be nice if we had a test like: https://primer.style/primitives/storybook/?path=/story/testing-colors--display-demo&globals=theme:light_high_contrast

Sadly I can not add copilot to your PR to do work which I guess is a good security thing.

Also why don't we use: `--display-[color]-borderColor-emphasis` instead of fgColor? And I am also wondering if we even need `label` colors if display colors doe exactly the same?

PR #1239, opened 2025-07-16
Update `control` knob border-color values

langermank:
Closes https://github.com/github/primer/issues/5466

Updates the `knob` border-color to align with our decision to only offer high contrast borders within high contrast themes.

Before/after


lukasoppermann (approved):
@langermank ~I think the toggle does not look that great because the knob border and outside border is the same.
I think we could either make it look like segmented control and have the border overlap the outside border. Or we could remove it entirely and have no border. But this way looks a big buggy to me.
The other components look great.~

Nevermind, if you look at it at 100% and not zoomed in, this is not really noticable.

PR #1242, opened 2025-08-01
Add progress bar tokens

langermank:
Adds `progressBar` tokens to support high contrast themes. Default themes will respect the existing design, while high contrast themes will include custom item colors and an overall track outline.

Check out Primer React for a visual: https://primer-6385406b5d-13348165.drafts.github.io/storybook/?path=/story/components-progressbar-features--all-colors&globals=colorScheme:all

High contrast themes


llastflowers (approved):
✅

PR #1245, opened 2025-08-07
Don't export new high contrast modes

lukasoppermann:
This PR make sure that we only export indented color themes for Figma.


langermank (approved):
This is really nice, I like how there's a prop so we can quickly change our minds later if we want 😄

PR #1268, opened 2025-10-30
Release npm trusted providers

jonrohan:
Part of https://github.com/github/primer/issues/6015

I also have a companion PR that will need to ship after this change to fix the integration workflows https://github.com/github/github-ui/pull/5972

This updates our workflows to allow for the NPM trusted publishing. With this we will no longer need to use tokens to publish the packages. https://docs.npmjs.com/trusted-publishers

One caveat though is that the configuration only allows for one workflow to publish. Because of this I needed to combine release.yml, release_candidate.yml, release_canary.yml into 1 file.

### Testing & Reviewing

I tested the canary release obviously here, I'm not sure we can test the main and candidate scenario without merging and fixing any issues that might exist. So in review, take a closer look at the `release-main:` and `release-candidate:` jobs.

### DIFF
--- a/.github/actions/upload-versions/action.yml
+++ b/.github/actions/upload-versions/action.yml
@@ -3,10 +3,12 @@ description: 'Upload version information for public packages'
 runs:
   using: 'composite'
   steps:
-    - name: Set up Node.js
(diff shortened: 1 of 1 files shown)


langermank (approved):
🚀

PR #1351, opened 2026-04-15
fix: preserve CSS variable references in border and outline tokens

lukasoppermann:
## Summary

Border and outline tokens were resolving to raw values instead of keeping CSS variable references.

**Before:**
```css
--border-default: 0.0625rem solid #d1d9e0;
--focus-outline: 2px solid #0969da;
```

**After:**
```css
--border-default: var(--borderWidth-default) solid var(--borderColor-default);
--focus-outline: var(--focus-outline-width) solid var(--focus-outline-color);
```

## Root cause

Style Dictionary's `outputReferencesTransformed()` always returns `false` for composite tokens (original value is an object, not a string), and `outputReferencesFilter()` returns `false` because width tokens live in a separate CSS file. Together, these prevented reference output for all composite border tokens.

## Fix

Added a `shouldOutputReferences()` helper in `src/platforms/css.ts` that returns `true` for `$type === 'border'` tokens. Style Dictionary's formatter already correctly handles `var()` substitution in flattened composite strings — it just needed to be enabled.

## Validation

- All lint, format, tests, and build pass
- Verified output across light, dark, and dark-dimmed themes
- Alias border tokens (e.g. `border.closed.emphasis`) continue to work correctly

### DIFF
--- a/src/platforms/css.ts
+++ b/src/platforms/css.ts
@@ -3,6 +3,28 @@ import type {PlatformInitializer} from '../types/platformInitializer.js'
 import type {PlatformConfig, TransformedToken} from 'style-dictionary/types'
 import {outputReferencesTransformed, outputReferencesFilter} from 'style-dictionary/utils'
 
+/**
+ * Determines whether a token's value should use CSS variable references.
+ *
+ * For most tokens, we check both that referenced tokens are in the current output
+ * (outputReferencesFilter) and that the string replacement is valid (outputReferencesTransformed).
+ *
+ * Composite border tokens need special handling because:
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey @jonrohan this PR keeps token references for borders, etc. this however means that you need to always import border.css with the themes if you use borders (or you import primer.css which has everything). Will this be an issue for us?

jonrohan:
I think we should be fine, we import the themes from `@import '@primer/primitives/dist/css/functional/themes/light.css';` for each theme.

jonrohan (approved):

PR #1375, opened 2026-05-05
chore: prepare pre-commit hook for husky v10

lukasoppermann:
Remove deprecated husky v9 lines that will fail in v10.0.0

## Changes
- Removed `. "$(dirname -- "$0")/_/husky.sh"` initialization line

These lines are marked as deprecated in husky v9 and will cause failures when husky v10 is released. The hook remains fully functional with only the actual implementation retained.

## Testing
- Verified with full test suite: lint, format, test, and build all pass
- Pre-commit hook executes the removed token check as intended


lukasoppermann:
@copilot Only keep one file: [pre-commit](https://github.com/primer/primitives/pull/1375/changes#diff-d2bc4bbf14eadc292d84e0ef5f7a93115c23557f533809fc80b896934291529d) in this PR

PR #1388, opened 2026-05-11
Fix formatting of storybook:install script in package.json

### LINKED ISSUE #1386: Recently added post install hook is broken when installing the published package from npm
### Describe the bug

https://github.com/primer/primitives/pull/1377

This copilot assisted PR slopped and broke the published package on npm. For example: https://github.com/buttons/github-buttons/actions/runs/25645985805/job/75275353725#step:4:16

### Reproduction steps

```bash
npm install @primer/primitives@11.8.0 
```

```
npm error code 2
npm error path /home/runner/work/github-buttons/github-buttons/node_modules/@primer/primitives
npm error command failed
npm error command sh -c npm run storybook:install
npm error > @primer/primitives@11.8.0 storybook:install
npm error > cd docs/storybook && npm ci --no-audit --no-fund
npm error sh: 1: cd: can't cd to docs/storybook
```

### Expected behavior

It should just work.

joshblack:
Closes #1386 

Remove the `postinstall` hook as this causes issues when using the package. This is only for local development and should not be included in the published package.

### DIFF
--- a/.changeset/funny-foxes-deliver.md
+++ b/.changeset/funny-foxes-deliver.md
@@ -1,0 +1,5 @@
+---
+"@primer/primitives": patch
+---
+
+Remove postinstall script from package.json
--- a/package.json
+++ b/package.json
@@ -47,8 +47,7 @@
     "prepack": "npm run build",
     "release": "changeset publish",
     "storybook": "npm run build:tokens && cd docs/storybook && npm run storybook",
-    "storybook:install": "cd docs/storybook && npm ci --no-audit --no-fund",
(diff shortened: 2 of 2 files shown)


joshblack:
FYI @lukasoppermann 

I think the project can be updated to use npm workspaces so that dependencies are setup automatically if that helps

francinelucca (approved):

lukasoppermann:
@joshblack I tried it and ran into issues. If you can get it to work, feel free to do so, that would be nice.

